PR #14, opened 2020-04-20
Adds Voice Connection support

matthewpi:
Does what the title says, however this implementation only enables the sending of Opus packets and does not handle the receiving of them, not like many people would want to receive them anyways.  I plan to try and add integration tests and an example that is able to play `.ogg` files encoded with opus.


matthewpi:
I believe this is ready to merge on my part, I have done a decent amount of testing and made an example bot (not in this repository because it has extra dependencies that I didn't want to add).

https://github.com/matthewpi/ayaya is the example bot

PR #26, opened 2020-05-07
fix(api): no retries on error

### LINKED ISSUE #25: fix(api): no retries on error
Regarding: `httputil.Client.Request`

Currently, `Request` retries on any HTTP Response status that is `< 200` or `> 299`.
However, many errors won't get resolved by a retry, as only 5xx errors indicate a server error and any other status indicates either an error with the data that was sent, etc..
An exception to this is of course code `429 - too many requests` which indicates rate limiting.
See also https://en.wikipedia.org/wiki/List_of_HTTP_status_codes.

I would propose to check for server or rate limit errors and retry only then, but let all others break the loop and get handled respectively.
This could be done by simply replacing the
```go
if status = r.GetStatus(); status < 200 || status > 299 {
    continue
}
```
with
```go
if status = r.GetStatus(); status == http.StatusTooManyRequests || status >= 500 {
    continue
}
```
.

If you create a branch for it, I can hand in a PR for it.

mavolin:
Closes #25.

### DIFF
--- a/utils/httputil/client.go
+++ b/utils/httputil/client.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"io"
 	"mime/multipart"
+	"net/http"
 
 	"github.com/diamondburned/arikawa/utils/httputil/httpdriver"
 	"github.com/diamondburned/arikawa/utils/json"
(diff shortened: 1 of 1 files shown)


mavolin:
> Can you swap in StatusTooManyRequest for a new constant with the same name in the same package?

Exposed or unexposed?

PR #28, opened 2020-05-07
API: fix #27

### LINKED ISSUE #27: API: payload for ModifyEmoji always empty
The payload sent by `api.Client.ModifyEmoji` is always empty, as the param struct is initialized with zero values, but the values passed to the method never get assigned to the payload.

I will create a PR fixing this.

mavolin:
Closes #27.

### DIFF
--- a/api/emoji.go
+++ b/api/emoji.go
@@ -1,8 +1,6 @@
 package api
 
 import (
-	"errors"
-
 	"github.com/diamondburned/arikawa/discord"
 	"github.com/diamondburned/arikawa/utils/httputil"
 )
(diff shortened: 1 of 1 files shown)


mavolin:
I also found that the `ErrEmojiTooLarge` variable in the same file is unused. Should I create a commit removing that?

diamondburned:
> I also found that the `ErrEmojiTooLarge` variable in the same file is unused. Should I create a commit removing that?

Yes, that should be removed. I suppose it was meant to be used with `image.Validate()`, but that already has its own error.

mavolin:
Done. Could you, if possible, release a hotfix release for this, as I need this method for a project I am working on.

diamondburned:
Will do as soon as CI passes.

PR #40, opened 2020-05-11
29 option types

mavolin:


api/channel.go
```diff
@@ -20,15 +20,15 @@ type CreateChannelData struct {
 	Name  string `json:"name"` // 2-100
 	Topic string `json:"topic,omitempty"`
 
-	Type discord.ChannelType `json:"type,omitempty"`
+	Type *discord.ChannelType `json:"type,omitempty"`
```
diamondburned, line 23: This doesn't need to be a pointer: 0, the zero-value, is also the [default value](https://discord.com/developers/docs/resources/channel#channel-object-channel-types). Perhaps omitempty could be omitted.

discord/channel.go
```diff
@@ -58,14 +58,14 @@ func (ch Channel) IconURL() string {
...
 
-const (
-	GuildText ChannelType = iota
-	DirectMessage
-	GuildVoice
-	GroupDM
-	GuildCategory
-	GuildNews
-	GuildStore
+var (
```
diamondburned, line 61: Why did we need this again?

discord/guild_const.go
```diff
@@ -49,38 +53,85 @@ const (
...
 
-type ExplicitFilter uint8
+// ExplicitFilter is the explicit content filter level of a guild.
+type ExplicitFilter nullable.Enum
 
-const (
-	NoContentFilter ExplicitFilter = iota
-	MembersWithoutRoles
-	AllMembers
+var (
```
diamondburned, line 59: Above structs already use `*ExplicitFilter`, do we need a `nullable.Enum` here? Also why var?

discord/guild_const.go
```diff
@@ -49,38 +53,85 @@ const (
...
+	// NoContentFilter disables content filtering for the guild.
+	NoContentFilter ExplicitFilter = 0
+	// MembersWithoutRoles filters only members without roles.
+	MembersWithoutRoles ExplicitFilter = 1
+	// AllMembers enables content filtering for all members.
+	AllMembers ExplicitFilter = 2
 )
 
-type Notification uint8
+func (f *ExplicitFilter) UnmarshalJSON(b []byte) error {
```
diamondburned, line 71: (If we're keeping this), `ExplicitFilter` should've already inherited `UnmarshalJSON` from `nullable.Enum` right?

discord/guild_const.go
```diff
@@ -49,38 +53,85 @@ const (
...
-	NoVerification Verification = iota
+	return err
+}
+
+func (n Notification) MarshalJSON() ([]byte, error) { return nullable.EnumToJSON(nullable.Enum(n)), nil }
+
+// Verification is the verification level required for a guild.
+type Verification nullable.Enum
+
+var (
```
diamondburned, line 107: Refer to above comment.

utils/json/nullable/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package nullable provides nullable types that get serialized to JSON null.
```
diamondburned, line 1: This file should be `nullable.go`.

utils/json/nullable/enum.go
```diff
@@ -0,0 +1,34 @@
+package nullable
+
+import "strconv"
+
+// EnumNull is the value used to represent JSON null.
+// It should never be used as a value, as it won't get serialized as such.
+const EnumNull = -1
+
+// Enum is a nullable version of a uint8.
+// Enum values should only consist of positive values, as negative values are reserved for internal constants, such as
```
diamondburned, line 10: (((nitpicking))) Can this comment be limited to 80 columns?

diamondburned:
Wait I clicked the circle recycling button next to my name under Reviewers. I don't know what happened.

discord/channel.go
mavolin: So we can write something like this:

```go
SomeStruct{
    Type: &GuildText, // type of Type is *ChanneType
}
```

If this were a constant, we couldn't make it a pointer, as it is already immutable.

discord/guild_const.go
mavolin: [Docs](https://discord.com/developers/docs/resources/guild#modify-guild) say the field is nullable, which I assume triggers a reset to default.

discord/guild_const.go
mavolin: Thought that as well, but unfortunately no. Inheritance seems to be the only thing for structs but not for type wrapping, so we have to call the function ourselves.
And as we can't use them directly, I decided to simplify things for the caller and don't implement the `Unmarshaler` and `Marshaler` interfaces on them, but to make a function for this, as this shortens UnmarshalJSON and MarshalJSON.

discord/guild_const.go
mavolin: For var see above comment

discord/guild_const.go
mavolin: Again the `ModifyGuild` payload allows this type to either be `null` or omitted.

api/channel.go
mavolin: If it's the default, we might as well omit it. Omitting it also saves us a few bytes in the JSON payload, so I'd say we keep it.

mavolin: But pointer removed

utils/json/nullable/doc.go
mavolin: Done. Also did `option/doc.go` -> `option/option.go`

mavolin:
Means you added yourself as reviewer.

utils/json/nullable/enum.go
mavolin: sure, should I refactor the rest of the code as well to use 80 cols?

mavolin: or is this just for comments?

mavolin:
> I haven't reviewed all of this yet, but here's a bigger question: could we instead do struct { Value T, Init bool }?

For what? For the option and nullable types? This would mean they can't get dereferenced that easily so Idk if it's worth it.

utils/json/nullable/enum.go
diamondburned: Code is 100 cols, comments is 80 cols is the convention.

diamondburned:
> For what? For the option and nullable types? This would mean they can't get dereferenced that easily so Idk if it's worth it.

It means you would be able to avoid pointers completely. Well, ideally anyway. That's how Option<T> works in most other languages.

utils/json/nullable/enum.go
mavolin: k

mavolin:
I think we should keep it this way. Writing a struct for every option type, especially in `ModifyXData` structs would result in super long structs, that just aren't readable, considering using the proposed struct takes up 3 extra lines.

diamondburned:
> I think we should keep it this way. Writing a struct for every option type, especially in `ModifyXData` structs would result in super long structs, that just aren't readable, considering using the proposed struct takes up 3 extra lines. And even with a constructor method, the struct variant
> a) takes up more memory.
> b) harder to access.

It's only making option structs for primitives. For example:

```go
package option

// type Enum *uint8
type Enum struct {
	Value uint8
	Init  bool
}
```

Also, sometimes we could sacrifice memory: a struct like this is cache friendlier than traversing a pointer.

diamondburned:
I could make another issue for this and merge this PR right now, if you want.

mavolin:
No that's alright.

mavolin:
If we decide otherwise merging this wouldn't make much sense.

mavolin:
I'd say we keep it, as I don't see any major advantage over the current implementation. And as `XData` structs normally don't have that long of a 'lifespan', seeing as they are just sent out and the gc'ed, I don't see those caching advantages.

PR #50, opened 2020-05-12
API: Deprecate FormatEmojiAPI in favor of a simpler function

### LINKED ISSUE #42: API: Deprecate FormatEmojiAPI in favor of a simpler function
`FormatEmojiAPI` takes two parameters, the first one being optional, which is not very idiomatic.
Also `FormatEmojiAPI` doesn't even return a `EmojiAPI` type, but a `string`.
Also, if we're doing this `APIEmoji` might be a better name, since this is an emoji used by the API and not an API for emojis.
Therefore, I propose to either deprecate or delete `FormatEmojiAPI` in favor of two new functions:

```go
func NewAPIEmojiFromGuildEmoji(id discord.Snowflake, name string) APIEmoji {
    return name+":"+id.String()
}

func NewAPIEmojiFromUnicode(emoji string) APIEmoji {
    return name
}
```

Additionally, `APIEmoji` should get a `URLEncode() string` method, that URL encodes the emoji for safe API usage.

Can hand in a PR for this

mavolin:
Closes #42.

### DIFF
--- a/api/emoji.go
+++ b/api/emoji.go
@@ -8,10 +8,10 @@ import (
 // Emoji is the API format of a regular Emoji, both Unicode or custom.
 type Emoji = string
 
-// NewEmoji creates a new Emoji using a custom guild emoji as
+// NewCustomEmoji creates a new Emoji using a custom guild emoji as
 // base.
 // Unicode emojis should be directly passed to the function using Emoji.
-func NewEmoji(id discord.Snowflake, name string) Emoji {
+func NewCustomEmoji(id discord.Snowflake, name string) Emoji {
 	return name + ":" + id.String()
 }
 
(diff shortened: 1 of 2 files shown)


api/emoji.go
```diff
@@ -5,14 +5,13 @@ import (
...
-	if id == 0 {
-		return name
-	}
+// Emoji is the API format of a regular Emoji, both Unicode or custom.
+type Emoji = string
 
+// NewEmoji creates a new Emoji using a custom guild emoji as
+// base.
+// Unicode emojis should be directly passed to the function using Emoji.
+func NewEmoji(id discord.Snowflake, name string) Emoji {
```
diamondburned, line 14: This should be called NewCustomEmoji imho.

PR #57, opened 2020-05-12
API: create data structs for various API methods

### LINKED ISSUE #44: API: create data structs for various API methods
Some API methods just take a lot of parameters and build a payload with that. However, this bears two problems:

1. Method signatures get very long, and sometimes even invoking them becomes a two-line job
2. Often some of the parameters are omittable. As go doesn't support function overloads, we can't create a bunch of methods for this that would cover every case, but we can use structs

### Proposal

1. All methods, that take three or more data arguments or have more than five total arguments, should use a `XData` struct.
2. Methods, where the API allows omittable fields, should, except for a few sane exceptions, always use a `XData` struct, to prevent unidiomatic and ugly zero-values in function parameters.

Can hand in a PR this.

mavolin:
Closes #44.

Also, adds documentation for ALL API function/types/constants as well as some improvements for the option types.


diamondburned:
Impressive!

diamondburned:
There are merge conflicts, can that be fixed?

mavolin:
dont push yet

mavolin:
Merge Conflicts fixed.

PR #58, opened 2020-05-12
Gateway: fix missing inheritance

### LINKED ISSUE #45: Gateway: fix missing inheritance
Some gateway events wrap a discord type and not inherit it in a struct:

```go
type UserUpdateEvent discord.User
```
instead of 
```go
type UserUpdateEvent struct {
    discord.User
}
```

Just wrapping the type does inherit the methods of it's underlying type, but not it's methods. So if one would want to get the mention of the updated user, on couldn't call `userUpdate.Mention()` as it is not inherited.
However, using a struct with `discord.User` as anonymous field, would allow `UserUpdateEvent` to inherit `discord.User`'s fields but also it's methods.

Can hand in a PR for this.

mavolin:
Closes #45.

### DIFF
--- a/gateway/events.go
+++ b/gateway/events.go
@@ -230,7 +230,9 @@ type (
 // https://discordapp.com/developers/docs/topics/gateway#presence
 type (
 	// Clients may only update their game status 5 times per 20 seconds.
-	PresenceUpdateEvent   discord.Presence
+	PresenceUpdateEvent struct {
+		discord.Presence
+	}
 	PresencesReplaceEvent []discord.Presence
 
 	// SessionsReplaceEvent is an undocumented user event. It's likely used for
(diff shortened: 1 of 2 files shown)


diamondburned:
Does this really work? I feel like it might confuse the reflect-based event callback handler.

Personally, I don't have any issues with the methods not inheriting, as it's just a simple type conversion.

diamondburned:
It doesn't work: https://play.golang.org/p/wTwgpRUjGDy

mavolin:
This should work though.

diamondburned:
Did you check if `package state` compiles? I think it might need some modifications.

diamondburned:
```
# github.com/diamondburned/arikawa/state
state/state_events.go:227:34: cannot convert ev (type *gateway.PresenceUpdateEvent) to type *discord.Presence
state/state_events.go:246:61: cannot convert *ev (type gateway.UserGuildSettingsUpdateEvent) to type gateway.UserGuildSettings
state/state_events.go:251:45: cannot convert ev (type *gateway.UserSettingsUpdateEvent) to type *gateway.UserSettings
state/state_events.go:257:46: cannot convert ev (type *gateway.UserUpdateEvent) to type *discord.User
state/state_events.go:262:30: cannot convert ev (type *gateway.VoiceStateUpdateEvent) to type *discord.VoiceState
```

PR #66, opened 2020-05-15
API: add EditText and EditEmbed methods

### LINKED ISSUE #63: API: add EditText and EditEmbed methods
Create edit methods for the `SendText` and `SendEmbed` methods.

Can hand in a PR.

mavolin:
Closes #63.

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -186,11 +186,11 @@ func (c *Client) EditEmbed(
 // EditMessageComplex.
 func (c *Client) EditMessage(
 	channelID, messageID discord.Snowflake, content string,
-	embed discord.Embed, suppressEmbeds bool) (*discord.Message, error) {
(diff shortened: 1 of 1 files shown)


api/message.go
```diff
@@ -161,15 +162,35 @@ type EditMessageData struct {
...
+		Embed: &embed,
+	})
+}
+
+// EditMessage edits a previously sent message. For more documentation, refer to
 // EditMessageComplex.
 func (c *Client) EditMessage(
 	channelID, messageID discord.Snowflake, content string,
-	embed *discord.Embed, suppressEmbeds bool) (*discord.Message, error) {
+	embed discord.Embed, suppressEmbeds bool) (*discord.Message, error) {
```
diamondburned, line 189: Can this be kept as a pointer? This will break quite a few APIs, and making it a pointer allows for optional.

mavolin: Of course, I didn't think that one through.

PR #79, opened 2020-05-16
State: Parallelize API requests in abstract functions

### LINKED ISSUE #76: State: add FetchSync option
When requesting a resource from the state, it attempts to fetch it from the store, and if not found there starts a request against Discord's API. This quickly becomes a problem of efficiency for methods such as `Permissions`, which, in theory, could trigger up to 3 API requests consecutively, which costs a lot of time.
To improve speed I propose to add a `FetchSync` field to the `State` struct (this could also be named `FetchAsync`, but I think speed is the preferred behavior in most cases, hence default to async behavior). This means a default state will fetch all field asynchronously, starting a goroutine for each individual request, if there are more than one resources requested within a state method. Only if explicitly set to `true` the old, slower behavior will be applied.

Alternatively, this could be made default, without a way to change it.

If approved, I can hand in a PR for this.

mavolin:
Closes #76.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -122,14 +122,14 @@ func (s *State) MemberDisplayName(guildID, userID discord.Snowflake) (string, er
 }
 
 func (s *State) AuthorColor(message *gateway.MessageCreateEvent) (discord.Color, error) {
-	if !message.GuildID.Valid() {
-		return 0, nil
+	if !message.GuildID.Valid() { // this is a dm
+		return discord.DefaultMemberColor, nil
 	}
 
 	if message.Member != nil {
 		guild, err := s.Guild(message.GuildID)
 		if err != nil {
-			return 0, nil
+			return 0, err
 		}
 		return discord.MemberColor(*guild, *message.Member), nil
 	}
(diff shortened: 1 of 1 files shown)


state/state.go
```diff
@@ -121,34 +121,49 @@ func (s *State) MemberDisplayName(guildID, userID discord.Snowflake) (string, er
 	return member.Nick, nil
 }
 
-func (s *State) AuthorColor(message *gateway.MessageCreateEvent) discord.Color {
+func (s *State) AuthorColor(message *gateway.MessageCreateEvent) (discord.Color, error) {
 	if !message.GuildID.Valid() {
-		return discord.DefaultMemberColor
+		return 0, nil
```
diamondburned, line 126: Is DefaultMemberColor removed?

mavolin: It's still used in discord.MemberColor

diamondburned: But this could return a 0 still, so why not make it return default?

mavolin: Let me fix that.

PR #116, opened 2020-06-06
Gateway: Split GuildCreateEvent

### LINKED ISSUE #113: Gateway: Split GuildCreateEvent
Split the GuildCreateEvent into three separate state events, that match the scenarios described in the docs:
> 1. When a user is initially connecting, to lazily load and backfill information for all unavailable guilds sent in the Ready event.
> 2. When a Guild becomes available again to the client.
> 3. When the current user joins a new Guild.

~ https://discord.com/developers/docs/topics/gateway#guild-create

Possible names:
1. `GuildReady` - this event also gets fired for every guild when connecting.
2. `GuildAvailable`
3. `GuildJoin`

The current `GuildCreateEvent` remains as is to ensure backward compatibility and acts as a wrapper for all three events.

Furthermore, the `GuildDelete` event should also be split up into `GuildLeave` and `GuildUnavailable`.

Approach:
- Store all Guilds ids sent in an Unavailable event in a set
- If `GuildCreateEvent` is fired, check if it's there
- If so, it's a `GuildAvailable` event
- If not check if we know this guild, i.e. it's in the state
- If so, it's a late `GuildReady` event
- If not, it's a `GuildJoin` event

mavolin:
Closes #113.

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -75,6 +75,34 @@ func (h *Handler) Call(ev interface{}) {
 	}
 }
 
+// CallDirect is the same as Call, but only calls those event handlers that
+// listen for this specific event, i.e. that aren't interface handlers.
+func (h *Handler) CallDirect(ev interface{}) {
+	var evV = reflect.ValueOf(ev)
+	var evT = evV.Type()
+
+	h.hmutex.RLock()
+	defer h.hmutex.RUnlock()
+
+	for _, order := range h.horders {
+		handler, ok := h.handlers[order]
+		if !ok {
+			// This shouldn't ever happen, but we're adding this just in case.
+			continue
+		}
+
+		if evT != handler.event {
+			continue
+		}
+
+		if h.Synchronous {
+			handler.call(evV)
+		} else {
+			go handler.call(evV)
+		}
+	}
+}
+
 // WaitFor blocks until there's an event. It's advised to use ChanFor instead,
 // as WaitFor may skip some events if it's not ran fast enough after the event
 // arrived.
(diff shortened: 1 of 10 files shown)


utils/moreatomic/snowflake_set.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// Delete deletes the passed discord.Snowflake from the set and returns true if
+// the element is present. If not, Delete is a no-op and returns false.
+func (s *SnowflakeSet) Delete(flake discord.Snowflake) bool {
+	s.mut.Lock()
+	defer s.mut.Unlock()
+
+	if _, ok := s.set[flake]; ok {
+		delete(s.set, flake)
```
diamondburned, line 46: Delete is already a noop without a check

handler/handler.go
```diff
@@ -75,6 +75,34 @@ func (h *Handler) Call(ev interface{}) {
 	}
 }
 
+// CallDirect is the same as Call, but only calls those event handlers that
+// listen for this specific event, i.e. that aren't interface handlers.
+func (h *Handler) CallDirect(ev interface{}) {
```
diamondburned, line 80: Why do we need this?

session/session.go
```diff
@@ -38,22 +28,27 @@ type Session struct {
...
 	hstop chan struct{}
+
+	// unavailableGuilds is a set of discord.Snowflakes of guilds that became
+	// unavailable when already connected to the gateway, i.e. sent in a
+	// GuildUnavailableEvent.
+	unavailableGuilds *moreatomic.SnowflakeSet
+	// unreadyGuilds is a set of discord.Snowflakes of guilds that were
+	// unavailable when connecting to the gateway, i.e. they had Unavailable
+	// set to true during Ready.
+	unreadyGuilds *moreatomic.SnowflakeSet
```
diamondburned, line 39: Can this be moved to state? This does count as storing something, so session is a bad place to put it.

utils/moreatomic/snowflake_set.go
mavolin: I know, but if I deleted the entry before the if, we wouldn't know if the entry is in the set. And we need this feedback to save ourselves mutexes in the `Session` that have to be manually locked/unlocked

handler/handler.go
mavolin: Without this, subscribers to `interface{}` events would both get e.g. a `GuildCreateEvent` and a `GuildJoinEvent`, effectively giving them the same event twice. This way only the `GuildJoinEvent` gets sent, as, in contrast to the `GuildCreateEvent` it gives situational information. `GuildCreateEvent`s can still be subscribed to, but directly, i.e. `AddHandler(func (*gateway.GuildCreateEvent))`.

diamondburned: I think those handlers should be aware of this. This is mostly for event handlers that do its own event type switching, so we can't miss out a few.

PR #134, opened 2020-07-29
Linting and typo fixes

### LINKED ISSUE #126: v1: Lint and docs
As a final step before creating the v1 release, everything should be

- [ ] linted
- [ ] major exposed functions/types should have a godoc

mavolin:
Closes #126.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -224,7 +224,7 @@ func (g *Gateway) ReconnectCtx(ctx context.Context) (err error) {
 	for i := 1; ; i++ {
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
+			return err
 		default:
 		}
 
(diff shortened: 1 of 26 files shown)


gateway/gateway.go
```diff
@@ -214,26 +214,36 @@ func (g *Gateway) Reconnect() {
...
 	wsutil.WSDebug("Reconnecting...")
 
 	// Guarantee the gateway is already closed. Ignore its error, as we're
 	// redialing anyway.
 	g.Close()
 
 	for i := 1; ; i++ {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
diamondburned, line 227: This should `return err` instead of `ctx.Err()`.

gateway/gateway.go
```diff
@@ -214,26 +214,36 @@ func (g *Gateway) Reconnect() {
...
 		wsutil.WSDebug("Trying to dial, attempt", i)
 
 		// Condition: err == ErrInvalidSession:
 		// If the connection is rate limited (documented behavior):
 		// https://discordapp.com/developers/docs/topics/gateway#rate-limiting
 
-		if err := g.OpenContext(ctx); err != nil {
-			return errors.Wrap(err, "failed to open gateway")
+		// make sure we don't overwrite our last error
+		if err2 := g.OpenContext(ctx); err2 != nil {
```
diamondburned, line 238: You can just do `if err = g.OpenContext(ctx); err != nil {` here.

PR #154, opened 2020-10-30
WS: reconnect timeout

mavolin:


gateway/gateway.go
```diff
@@ -185,18 +200,15 @@ func (g *Gateway) Close() error {
...
+	ctx, cancel := context.WithTimeout(context.Background(), g.WSTimeout)
+	defer cancel()
+
+	g.ReconnectCtx(ctx)
 }
 
 // ReconnectCtx attempts to reconnect until context expires. If context cannot
 // expire, then the gateway will try to reconnect forever.
-func (g *Gateway) ReconnectCtx(ctx context.Context) (err error) {
+func (g *Gateway) ReconnectCtx(ctx context.Context) {
```
diamondburned: ReconnectCtx should still return an error, I think. Having something that takes in a context and may timeout not returning an error is kind of counter-intuitive.

Perhaps have the `FatalErrorCallback` ran before returning err.

diamondburned: This section is missing a check for `ReconnectTimeout`. If the timeout is 0, then reconnection should continue inside the loop.

The code should be something like this:

```go
if g.ReconnectTimeout > 0 {
    tctx, cancel := context.WithTimeout(ctx, g.ReconnectTimeout)
    defer cancel()

    ctx = tctx
}
```

gateway/gateway.go
```diff
@@ -185,18 +200,15 @@ func (g *Gateway) Close() error {
...
-		if err := g.ReconnectCtx(context.Background()); err != nil {
-			g.ErrorLog(err)
-		} else {
-			return
-		}
-	}
+	ctx, cancel := context.WithTimeout(context.Background(), g.WSTimeout)
+	defer cancel()
+
+	g.ReconnectCtx(ctx)
```
diamondburned, line 215: `ReconnectCtx` should be called with a `context.Background()`, as we want to default to endlessly trying to reconnect.

gateway/integration_test.go
```diff
@@ -88,9 +88,7 @@ func TestIntegration(t *testing.T) {
 		ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
 		defer cancel()
 
-		if err := gateway.ReconnectCtx(ctx); err != nil {
-			t.Fatal("Unexpected error while reconnecting:", err)
-		}
+		gateway.ReconnectCtx(ctx)
```
diamondburned: This should have the error check restored because of above review comments.

gateway/gateway.go
```diff
@@ -217,7 +229,7 @@ func (g *Gateway) ReconnectCtx(ctx context.Context) (err error) {
 		// https://discordapp.com/developers/docs/topics/gateway#rate-limiting
 
 		// make sure we don't overwrite our last error
-		if err = g.OpenContext(ctx); err != nil {
+		if err := g.OpenContext(ctx); err != nil {
```
diamondburned: Restore this to `if err = g.OpenContext(ctx)...`.

diamondburned (changes requested):

PR #159, opened 2020-11-10
API: abstract away boundary limits for DeleteMessages

mavolin:


api/message.go
```diff
@@ -331,8 +336,37 @@ func (c *Client) DeleteMessage(channelID discord.ChannelID, messageID discord.Me
...
+	switch {
+	case len(messageIDs) == 0:
+		return nil
+	case len(messageIDs) == 1:
+		return c.DeleteMessage(channelID, messageIDs[0])
+	case len(messageIDs) <= maxMessageDeleteLimit: // Fast path
+		return c.deleteMessagesRaw(channelID, messageIDs)
+	}
+
+	for start := 0; start < len(messageIDs); start += 100 {
```
diamondburned: `start += 100` should be `start += maxMessageDeleteLimit`.

api/message.go
```diff
@@ -331,8 +336,37 @@ func (c *Client) DeleteMessage(channelID discord.ChannelID, messageID discord.Me
...
+		end := start + maxMessageDeleteLimit
```
diamondburned: We should consider just making a `min(i, j int)` function for all these abstracted functions.

~~tfw no generics~~

api/message.go
```diff
@@ -331,8 +336,37 @@ func (c *Client) DeleteMessage(channelID discord.ChannelID, messageID discord.Me
...
+		err := c.deleteMessagesRaw(channelID, messageIDs[start:end])
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *Client) deleteMessagesRaw(channelID discord.ChannelID, messageIDs []discord.MessageID) error {
```
diamondburned: Maybe use `deleteMessages` instead of the `Raw` suffix?

diamondburned (changes requested):

api/utils.go
```diff
@@ -0,0 +1,10 @@
+package api
```
diamondburned: `utils.go` should be better named as `math.go`.

diamondburned (changes requested):

PR #161, opened 2020-11-15
Add user_id to SpeakingData

tystuyfzand:
Undocumented field, discordgo uses it. Likely used to map speaking states in the client to user ids.

### DIFF
--- a/voice/voicegateway/commands.go
+++ b/voice/voicegateway/commands.go
@@ -105,7 +105,7 @@ const (
 )
 
 // OPCode 5
-// https://discordapp.com/developers/docs/topics/voice-connections#speaking-example-speaking-payload
+// https://discord.com/developers/docs/topics/voice-connections#speaking-example-speaking-payload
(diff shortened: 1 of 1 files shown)


voice/voicegateway/commands.go
```diff
@@ -105,11 +105,12 @@ const (
 )
 
 // OPCode 5
-// https://discord.com/developers/docs/topics/voice-connections#speaking-example-speaking-payload
+// https://discordapp.com/developers/docs/topics/voice-connections#speaking-example-speaking-payload
```
diamondburned: Why?

diamondburned (changes requested):

voice/voicegateway/commands.go
tystuyfzand: You know, I had it open on v1 and accidentally noticed the wrong branch, probably copied the wrong url over. I'll fix it

PR #162, opened 2020-11-16
Add support for voice gateway events

tystuyfzand:
This PR adds support for voice gateway events, including a couple undocumented/partially documented ones.

Now supported:

- Speaking update (Op 5)
- Client Connect (Op 12)
- Client Disconnect (Op 13)

Note: Bots do not receive all speaking events, only one per user in the time the user is in the channel. This appears to be intentional, and members in the Discord API server have confirmed it.


diamondburned:
This PR will also introduce minor refactors into certain parts that don't particularly have to do with the voice gateway itself, as part of the overall refactor.

PR #163, opened 2020-11-18
State: don't check store if resource is not tracked through intents

mavolin:


state/state.go
```diff
@@ -682,7 +748,7 @@ func (s *State) Roles(guildID discord.GuildID) ([]discord.Role, error) {
 
 func (s *State) fetchGuild(id discord.GuildID) (g *discord.Guild, err error) {
 	g, err = s.Session.Guild(id)
-	if err == nil {
+	if err == nil && s.Gateway.Identifier.Intents.Has(gateway.IntentGuilds) {
```
diamondburned: Maybe we should have a `(*Gateway).HasIntent()` helper for this.

PR #170, opened 2020-11-22
Discord: add missing Integration fields

mavolin:


discord/guild_consts.go
```diff
@@ -160,8 +160,9 @@ func (v Verification) MarshalJSON() ([]byte, error) { return enum.ToJSON(enum.En
 type Service string
 
 const (
-	Twitch  Service = "twitch"
-	YouTube Service = "youtube"
+	ServiceTwitch  Service = "twitch"
```
diamondburned: For consistency, I personally think this should either be reverted or swapped to `TwitchService` and similar.

diamondburned (changes requested):

discord/guild_consts.go
mavolin: I'm going to go with `TwitchService` then. Reverting doesn't really make sense, seeing as `discord.Discord` is not really expressive.

PR #173, opened 2020-11-25
httputil: add timeout

mavolin:


api/rate/rate.go
```diff
@@ -111,6 +115,10 @@ func (l *Limiter) Acquire(ctx context.Context, path string) error {
 	}
 
 	if sleep > 0 {
+		if deadline, ok := ctx.Deadline(); ok && time.Now().Add(sleep).After(deadline) {
```
diamondburned: Instead of recalculating the time, we should check the deadline against the time that we already have. That should be more accurate.

diamondburned (changes requested):

PR #174, opened 2020-11-29
Add voice receive capability to v2

tystuyfzand:
This adds a zero-allocation voice receive option - which is obviously not thread safe.

Some notes about the implementation vs what was discussed in Discord:

- Buffers are fixed to 1400, ideally voice data will NEVER be this big, and this ensures that there is never an issue. An extra few bytes isn't much.
- The for loop is retained on ReadPacket, as length 11 packets are seen as well at some points. This ensures we get a valid voice packet, instead of one without the expected data.

### DIFF
--- a/voice/udp/udp.go
+++ b/voice/udp/udp.go
@@ -17,6 +17,15 @@ var Dialer = net.Dialer{
 	Timeout: 10 * time.Second,
 }
 
+// Packet represents a voice packet. It is not thread-safe.
+type Packet struct {
+	SSRC      uint32
+	Sequence  uint16
+	Timestamp uint32
+	Type      []byte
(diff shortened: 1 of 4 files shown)


voice/udp/udp.go
```diff
@@ -36,6 +53,12 @@ type Connection struct {
 	sequence  uint16
 	timestamp uint32
 	nonce     [24]byte
+
+	// recv fields
+	recvNonce  [24]byte
+	recvBuf    []byte  // len 1400
+	recvOpus   []byte  // len 1400
+	recvPacket *Packet // uses recvBuf's backing array
```
diamondburned: This comment should've said "uses recvOpus' backing array".

PR #189, opened 2021-04-03
Add automatic BotData connection into Gateway

diamondburned:
This pull request adds `gateway.BotData` usage into the gateway constructor to update the rate limiter. This is useful when the bot is being restarted. It also made new constructors to ease sharing `Identifier` instances around, which is necessary for sharding.

**Note** that there is currently no straightforward way to make bots use the new constructors. The intention is that most bots don't need sharding, and since bots that do need sharding need additional legwork to boot up multiple gateway instances that it would need more code anyway. Bots can thus borrow the new constructors to easily share a common` Identifier`.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"net/http"
 	"net/url"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mavolin (dismissed):

gateway/gateway.go
```diff
@@ -58,10 +58,12 @@ type SessionStartLimit struct {
...
-	return g.URL, httputil.NewClient().RequestJSON(
-		&g, "GET",
-		EndpointGateway,
-	)
+	c := httputil.NewClient()
+	if err := c.RequestJSON(&g, "GET", EndpointGateway); err != nil {
+		return "", err
+	}
+
+	return g.URL, nil
```
mavolin, line 66: What exactly does splitting this up accomplish? Seeing as `BotData` is not a pointer, `g.URL` would be `""` in an error scenario as well, so in my head, this revision should perform exactly the same as the previous version.

diamondburned: It does perform exactly the same. I only did this to make the line shorter and to make the code more obvious.

mavolin: Ah, alright!

mavolin (approved):

PR #190, opened 2021-04-03
Gateway: Handling of 'Sharding Required', Proper Closures and General Cleanup

mavolin:
### Sharding
This PR adds support for handling the `4011` aka "Sharding Required" close code, to allow using `Gateway` for sharding. Handling is provided through a `OnShardingRequired` function inside the `Gateway` struct, called if Discord closes a connection with a `4011` error code. Additionally, the Gateway now won't continue to attempt to reconnect, but instead exit.

### `Gateway.Close`
`Gateway.Close` was adapted to behave like `.CloseGracefully` in an effort to ensure, that a user made call to `.Close` will end the WS connection, even if `Gateway` is in the process of reconnecting. Consequently, `.CloseGracefully` has been deprecated, and it has been advised to use `.Close` instead. Furthermore, `.Reconnect` and `.ReconnectCtx` were unexported, as user-triggered reconnections are now impossible.

### Other Improvements
* When attempting to reconnect, `Gateway.reconnectCtx` will now wait for an increasing delay between each attempt to connect, with a max delay of 60 seconds.
* `Gateway.ReconnectTimeout` was deprecated in favor of the new `Gateway.ReconnectAttempts`, to have more control over retries despite delays between tries

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -33,7 +33,8 @@ var (
 
 var (
 	ErrMissingForResume = errors.New("missing session ID or sequence for resuming")
-	ErrWSMaxTries       = errors.New("max tries reached")
+	ErrWSMaxTries       = errors.New(
+		"could not connect to the Discord gateway before reaching the timeout")
 )
 
 // BotData contains the GatewayURL as well as extra metadata on how to
@@ -253,7 +254,7 @@ func (g *Gateway) Reconnect() {
 
 	if g.ReconnectTimeout > 0 {
 		var cancel func()
-		ctx, cancel = context.WithTimeout(context.Background(), g.WSTimeout)
+		ctx, cancel = context.WithTimeout(context.Background(), g.ReconnectTimeout)
 
 		defer cancel()
 	}
@@ -272,7 +273,7 @@ func (g *Gateway) ReconnectCtx(ctx context.Context) (err error) {
 	// redialing anyway.
 	g.Close()
 
-	for i := 1; ; i++ {
+	for try := 1; ; try++ {
 		select {
 		case <-ctx.Done():
 			g.FatalErrorCallback(ErrWSMaxTries)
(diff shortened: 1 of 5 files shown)


diamondburned (approved):

gateway/gateway.go
```diff
@@ -116,12 +128,18 @@ type Gateway struct {
 	// Defaults to noop.
 	FatalErrorCallback func(err error)
 
+	// OnScalingRequired is the function called, if discord closes with error
+	// code 4011 aka Scaling Required.
+	OnScalingRequired func()
```
diamondburned, line 137: Is this only called on initial identification or on reconnects as well?

gateway/gateway.go
```diff
@@ -246,56 +283,62 @@ func (g *Gateway) SessionID() string {
 	return g.sessionID
 }
 
-// Reconnect tries to reconnect until the ReconnectTimeout is reached, or if
+// reconnect tries to reconnect until the ReconnectTimeout is reached, or if
```
diamondburned: This is breaking the API; why the change?

diamondburned: There are also changes on the contextful version of this method.

utils/wsutil/ws.go
```diff
@@ -157,50 +157,37 @@ func (ws *Websocket) SendCtx(ctx context.Context, b []byte) error {
...
-	return err
+
+	if graceful {
+		if gc, ok := ws.conn.(GracefulCloser); ok {
+			WSDebug("Conn: Closing gracefully")
+			return gc.CloseGracefully()
+		}
+
+		WSDebug("Conn: The Websocket's Connection does not support graceful closure. Closing normally instead.")
+		return ws.conn.Close()
```
diamondburned: This like can just not be here; execution will fall out into the `Close()` after it. The redundant debug logging is fine.

diamondburned (changes requested):

gateway/gateway.go
mavolin: Because `Reconnect` does not work anymore, since Closures, as discussed, invalidate the session ID.

gateway/gateway.go
mavolin: This is called whenever Discord closes the WS with 4011, not afterwards. On call, the Gateway will be closed, and the user should signal all other shards to disconnect, and reconnect with more shards.

gateway/gateway.go
diamondburned: Closes aren't specifically relevant in the context of the caller; the function should automatically handle closing (ungracefully) and resume as normally.

---

For the record, I still think there should be a way to close without invalidating the session ID. I think a new method, maybe `Terminate`, should be added. The old `CloseGracefully` can call `Close` but marked as deprecated.

I'm not for breaking the API without bumping the major version.

utils/wsutil/ws.go
mavolin: I don't understand, please elaborate.

gateway/gateway.go
mavolin: Alright, will implemt this tomorrow.

mavolin: However, I think `Pause` would be more self-explanatory, as it emphasizes that the connection can be resumed, which not be so apparent when using `Terminate`. What do you think?

utils/wsutil/ws.go
diamondburned: Like this:

```go
	if graceful {
		if gc, ok := ws.conn.(GracefulCloser); ok {
			WSDebug("Conn: closing gracefully")
			return gc.CloseGracefully()
		}

		WSDebug("Conn: graceful closure unsupported; closing normally instead.")
	}

	WSDebug("Conn: closing")
	return ws.conn.Close()
```

mavolin: Sure, I just wanted to prevent double logging, but I guess `WSDebug` is verbose enough as for this to not matter. Again, I will do this tomorrow

gateway/gateway.go
diamondburned: Sounds good to me.

diamondburned (approved):

PR #191, opened 2021-04-04
v3: Gateway/Shard: Add Shard Manager

mavolin:


gateway/shard/manager.go
```diff
@@ -0,0 +1,327 @@
...
+	// the default when the given time is reached.
+	id.IdentifyGlobalLimit.SetBurst(botData.StartLimit.Remaining)
+	id.IdentifyGlobalLimit.SetBurstAt(resetAt, botData.StartLimit.Total)
+
+	// Update the maximum number of identify requests allowed per 5s.
+	id.IdentifyShortLimit.SetBurst(botData.StartLimit.MaxConcurrency)
+
+	gatewayURL := botData.URL + "?" + param.Encode()
+
+	m := newIdentifiedManager(gatewayURL, id, botData.Shards, GenerateShardIDs(botData.Shards)...)
```
chanbakjsd: This code seems to be duplicated with the code below.

gateway/shard/manager.go
```diff
@@ -0,0 +1,327 @@
...
+func newIdentifiedManager(
+	gatewayURL string, id *gateway.Identifier, totalShards int, shardIDs ...int) *Manager {
+
+	gateways := make([]*gateway.Gateway, len(shardIDs))
+
+	for i, shardID := range shardIDs {
+		id.SetShard(shardID, totalShards)
+		idCp := *id
+
+		gateways[i] = gateway.NewCustomIdentifiedGateway(gatewayURL, &idCp)
```
chanbakjsd, line 104: `Shard` seems to be a pointer to an array internally (`*[2]int`) so copying like this will only copy the pointer and not the data itself.
This might cause some race condition.

gateway/shard/manager.go
```diff
@@ -0,0 +1,327 @@
...
+
+		return m
+	}
+
+	return m, nil
+}
+
+// NewIdentifiedManager creates a new Manager using the passed url and the
+// passed gateway.Identifier. The shard information stored on the passed
+// identifier will be ignored. Instead totalShards and shardIDs will be used.
```
chanbakjsd, line 77: It might be worth noting that `NewIdentifiedManager` does not handle OnShardingRequired in the docs as I think the Manager's docs for OnShardingRequired seems to imply that OnShardingRequired is always present somewhat.

gateway/shard/manager.go
```diff
@@ -0,0 +1,327 @@
...
+		f(g)
+	}
+}
+
+// ApplyError is the same as Apply, but the iterator function returns an error.
+// If such an error occurs, the error will be returned wrapped in an *Error.
+//
+// If all is set to true, ApplyError will apply the passed function to all
+// gateways. If a single error occurs, it will be returned as an *Error, if
+// multiple errors occur then they will be returned as *MultiError.
```
chanbakjsd, line 193: `all` is unimplemented.

gateway/shard/manager.go
```diff
@@ -0,0 +1,327 @@
...
+				}
+			}
+
+			if len(errs) == 1 {
+				return errs[0]
+			}
+
+			return errs
+		}
+	}
```
chanbakjsd: This code can use ApplyError instead.

gateway/shard/manager.go
```diff
@@ -0,0 +1,327 @@
...
+// UpdateStatus updates the status of all gateways handled by this Manager.
+//
+// If an error occurs, UpdateStatus will attempt to update the status of all
+// remaining gateways first, before returning. If multiple errors occur during
+// that process, a MultiError will be returned.
+func (m *Manager) UpdateStatus(d gateway.UpdateStatusData) error {
+	return m.ApplyError(func(g *gateway.Gateway) error { return g.UpdateStatus(d) }, true)
+}
+
+func (m *Manager) RequestGuildMembers(d gateway.RequestGuildMembersData) error {
```
chanbakjsd, line 320: Exported function does not have documentation.

internal/moreatomic/bool.go
```diff
@@ -17,3 +17,12 @@ func (b *Bool) Set(val bool) {
...
 	atomic.StoreUint32(&b.val, x)
 }
+
+func (b *Bool) CompareAndSwap(old bool) bool {
+	var oldN uint32 = 0
+	if old {
+		oldN = 1
+	}
+
+	return atomic.CompareAndSwapUint32(&b.val, oldN, oldN+1%2)
```
chanbakjsd:
```suggestion
	return atomic.CompareAndSwapUint32(&b.val, oldN, (oldN+1) % 2)
```
The remainder operation is done before the add operation similar to how multiplication is done before add.

gateway/shard/shard.go
```diff
@@ -0,0 +1,81 @@
+// Package shard provides sharding capabilities through Manager, found in every
+// session.Session.
+package shard
+
+import (
+	"errors"
+	"strings"
+)
+
+// GenerateShardIDs generates an array of ints of 0..(total-1).
```
chanbakjsd:
```suggestion
// GenerateShardIDs generates a slice of ints of [0, total).
```
Array means a different thing in Go.

gateway/shard/shard.go
```diff
@@ -0,0 +1,81 @@
...
+	// Source is the error itself
+	Source error
+}
+
+func (err *Error) Unwrap() error {
+	return err.Source
+}
+
+func (err *Error) Error() string {
+	panic("implement me")
```
chanbakjsd: Please implement this.

gateway/shard/shard.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return ids
+}
+
+// Error is the error returned if an action on a single shard fails.
+type Error struct {
+	// ShardID is the id of the shard that returned the error.
+	ShardID int
+	// Source is the error itself
```
chanbakjsd: Inconsistent full stop.

gateway/shard/shard.go
```diff
@@ -0,0 +1,81 @@
...
+func (err *Error) Error() string {
+	panic("implement me")
+}
+
+// MultiError combines multiple errors in a slice.
+type MultiError []error
+
+func (errs MultiError) Error() string {
+	var b strings.Builder
+	n := 24 + 2*len(errs)
```
chanbakjsd: 24 feels too magical. Maybe make `"multiple errors occurred:"` a const string and do `len` here instead.

internal/moreatomic/bool.go
chanbakjsd:
```suggestion
	return atomic.CompareAndSwapUint32(&b.val, oldN, (oldN+1) % 2)
```
The remainder operation is done before the addition operation similar to multiplication and addition.

session/session.go
```diff
@@ -47,19 +49,27 @@ func NewWithIntents(token string, intents ...gateway.Intents) (*Session, error)
 		return nil, errors.Wrap(err, "failed to connect to Gateway")
 	}
 
-	return NewWithGateway(g), nil
+	return NewWithGateways(g), nil
```
chanbakjsd: Is it really ideal to only start one Gateway when Intents are specified?

session/session.go
```diff
@@ -47,19 +49,27 @@ func NewWithIntents(token string, intents ...gateway.Intents) (*Session, error)
...
+	}
+
+	m, err := shard.NewManager(token)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to connect to Gateway")
+		return nil, err
 	}
 
-	return NewWithGateway(g), nil
+	return NewWithShardManager(m), err
```
chanbakjsd: `err` is guaranteed to be `nil` so just return `nil` here.

state/state.go
```diff
@@ -119,7 +120,24 @@ func NewWithStore(token string, cabinet store.Cabinet) (*State, error) {
...
+
+func newWithAutoRescale(s *session.Session, cabinet store.Cabinet) *State {
+	state := NewFromSession(s, cabinet)
+	state.ShardManager.OnShardingRequired = func() *shard.Manager {
+		state.Cabinet.Reset()
+		token := s.ShardManager.Gateways()[0].Identifier.Token
+
+		m, err := shard.NewManager(token)
+		if err != nil {
+			return nil
```
chanbakjsd, line 138: Return an error in this function too instead of ignoring the error.

gateway/shard/shard.go
chanbakjsd:
```suggestion
// GenerateShardIDs generates a slice of integers of [0, total).
```
Arrays mean a different thing in Go.

gateway/shard/shard.go
chanbakjsd: Please implement this.

gateway/shard/shard.go
chanbakjsd: 24 seems like too magical of a number. Consider making the string an unexported const and `len` it instead.

internal/moreatomic/bool.go
chanbakjsd:
```suggestion
	return atomic.CompareAndSwapUint32(&b.val, oldN, (oldN+1) % 2)
```
The remainder operation is done before the addition operation similar to multiplication and addition.

state/state.go
```diff
@@ -666,14 +694,16 @@ func (s *State) Messages(channelID discord.ChannelID) ([]discord.Message, error)
 
 // Presence checks the state for user presences. If no guildID is given, it
 // will look for the presence in all cached guilds.
-func (s *State) Presence(gID discord.GuildID, uID discord.UserID) (*gateway.Presence, error) {
-	if !s.Gateway.HasIntents(gateway.IntentGuildPresences) {
+func (s *State) Presence(
+	guildID discord.GuildID, userID discord.UserID) (*gateway.Presence, error) {
+
```
chanbakjsd, line 739: Unnecessary indent change?

state/state.go
```diff
@@ -683,21 +713,22 @@ func (s *State) Presence(gID discord.GuildID, uID discord.UserID) (*gateway.Pres
...
-	return s.Cabinet.Presence(gID, uID)
+	return s.Cabinet.Presence(guildID, userID)
 }
 
 ////
 
-func (s *State) Role(guildID discord.GuildID, roleID discord.RoleID) (target *discord.Role, err error) {
-	if s.Gateway.HasIntents(gateway.IntentGuilds) {
+func (s *State) Role(guildID discord.GuildID, roleID discord.RoleID) (target *discord.Role,
+	err error) {
```
chanbakjsd: More unnecessary indent change?

state/state.go
```diff
@@ -753,17 +784,19 @@ func (s *State) Roles(guildID discord.GuildID) ([]discord.Role, error) {
...
 
 	return
 }
 
-func (s *State) fetchMember(gID discord.GuildID, uID discord.UserID) (m *discord.Member, err error) {
-	m, err = s.Session.Member(gID, uID)
-	if err == nil && s.Gateway.HasIntents(gateway.IntentGuildMembers) {
-		err = s.Cabinet.MemberSet(gID, *m)
+func (s *State) fetchMember(
+	guildID discord.GuildID, userID discord.UserID) (m *discord.Member, err error) {
```
chanbakjsd, line 836: Why is this pushed to the next line?

state/state.go
```diff
@@ -772,12 +805,14 @@ func (s *State) fetchMember(gID discord.GuildID, uID discord.UserID) (m *discord
 // tracksMessage reports whether the state would track the passed message and
 // messages from the same channel.
 func (s *State) tracksMessage(m *discord.Message) bool {
-	return (m.GuildID.IsValid() && s.Gateway.HasIntents(gateway.IntentGuildMessages)) ||
-		(!m.GuildID.IsValid() && s.Gateway.HasIntents(gateway.IntentDirectMessages))
+	return (m.GuildID.IsValid() &&
+		s.ShardManager.FromGuildID(m.GuildID).HasIntents(gateway.IntentGuildMessages)) ||
+		(!m.GuildID.IsValid() &&
+			s.ShardManager.FromGuildID(m.GuildID).HasIntents(gateway.IntentDirectMessages))
```
chanbakjsd: Might be worth lifting `s.ShardManager.FromGuildID(m.GuildID)` into a variable to maintain code similar to the old code.

state/state.go
```diff
@@ -772,12 +805,14 @@ func (s *State) fetchMember(gID discord.GuildID, uID discord.UserID) (m *discord
...
 }
 
 // tracksChannel reports whether the state would track the passed channel.
 func (s *State) tracksChannel(c *discord.Channel) bool {
-	return (c.GuildID.IsValid() && s.Gateway.HasIntents(gateway.IntentGuilds)) ||
-		!c.GuildID.IsValid()
+	return (c.GuildID.IsValid() && s.ShardManager.FromGuildID(c.GuildID).
+		HasIntents(gateway.IntentGuilds)) || !c.GuildID.IsValid()
```
chanbakjsd: Weird indentation.

chanbakjsd (changes requested):
Nice work. This PR adds shard manager to arikawa and moves `Session` (and by extension, `State` and `Bot` too) on top of it.
There are a few problems that needs to be addressed but otherwise this PR looks great!

gateway/shard/manager.go
mavolin: `OnShardingRequired` is actually never present unless using `NewManager`. So I'll try use better wording for both docs.

gateway/shard/manager.go
mavolin: It actually can't, because unlike the other methods like `Close`, `Open` attempts to close all previously opened gateways if it encounters an error.
This was done, because in most cases if you `Open` your session/state during startup, and encounter an error, you would most likely exit the program.

chanbakjsd: You can probably just loop over all gateways again to close them if `ApplyError` returned an error unless I'm missing something.

mavolin: No, because in an error scenario we would need the index of the gateway that returned an error, so we only close those gateways (i.e `m.gateways[:i]`, see line 248) that were previously opened.

state/state.go
mavolin: The function header was longer than the 100 character limit, so I wrapped it.

state/state.go
mavolin: The same here with the 100 character limit, but in this case it seems my IDE's formatter wrapped this incorrectly, so fixing this.

gateway/shard/manager.go
chanbakjsd: `ShardID` can provide that information. If the operation of searching based on ShardID is too expensive, it might be a sign that `Error` could use an "internal ID" to be more useful.

state/state.go
mavolin: Again, 100 character line limit

state/state.go
mavolin: This is the indention gofmt uses, what should I change? I could put the function call on the same line, but that would require an extra line.

chanbakjsd: I personally think another line doesn't hurt and will aid in readability instead.

gateway/shard/manager.go
mavolin: Sure, I can do that. Computational expense is negligible since: a) open is ever only performed once, b) network is the true bottleneck here, and c) there will never be a large number of shards stored in a manager.

state/state.go
mavolin: The idea here is, that if such an error occurs, all connections are closed. If the user wants to handle that error, e.g. log it, implement a more complex retry mechanism, they would have to provide a custom implementation.
I did this, because adding another logging/capture function that will ever just receive a single error seems way to bloated, especially since the above solution is just as elegant.

Edit: If it wasn't apparent, the connections are closed if `nil` is returned.

gateway/shard/manager.go
```diff
@@ -206,16 +194,30 @@ func (m *Manager) ApplyError(f func(g *gateway.Gateway) error, all bool) error {
...
+			}
+
+			errs = append(errs, wrapperErr)
 		}
 	}
 
-	return nil
+	if len(errs) == 0 {
+		return nil
+	} else if len(errs) == 1 {
```
chanbakjsd: Unnecessary `else` here. In fact, consider making it a `switch` on `len(errs)` instead.

gateway/shard/manager.go
chanbakjsd: I don't think copy pasting the documentation is the right move here. Especially considering it's going to get squashed into one row on pkg.go.dev.

chanbakjsd (changes requested):
The changes made had a few minor issues in it.

gateway/shard/manager.go
mavolin: Why is the else unnecessary? If we just encounter a single error, we should return just that one instead of the MultiError. Also we only have two cases, so why a switch?

chanbakjsd: The `else` is not required because the `if` contains an early return. A `switch` can contain `case 0:`, `case 1:` and `default:`.

gateway/shard/manager.go
mavolin: All docs for API related methods are copied from the official docs. The docs are usually the most extensive information we have, so if I were to write docs myself, I would basically just be rewording the docs found on the official API doc page.

state/state.go
chanbakjsd: Ah right. I was reading this as part of `newWithAutoRescale` instead of in `state.ShardManager.OnShardingRequired`. Feel free to resolve this.

gateway/shard/manager.go
```diff
@@ -0,0 +1,375 @@
...
+// in every of the Manager's gateways.
+func (m *Manager) onShardingRequired() {
+	if m.rescaleExec.CompareAndSwap(false) {
+		m.Close()
+
+		if m.Rescale == nil {
+			return
+		}
+
+		m.mutex.Lock()
```
diamondburned: This mutex should probably be moved to be acquired after `Rescale()`, just to prevent any potential deadlocks.

voice/voice.go
```diff
@@ -4,11 +4,16 @@
...
+	"github.com/diamondburned/arikawa/v2/gateway"
+	"github.com/diamondburned/arikawa/v2/gateway/shard"
+)
 
 // AddIntents adds the needed voice intents into gw. Bots should always call
 // this before Open if voice is required.
-func AddIntents(gw *gateway.Gateway) {
-	gw.AddIntents(gateway.IntentGuilds)
-	gw.AddIntents(gateway.IntentGuildVoiceStates)
+func AddIntents(m *shard.Manager) {
```
diamondburned: I think this should be split into a version for `Gateway` and another for the shard manager, just to make things just as easy as before for people implementing their own shard manager.

state/state.go
```diff
@@ -119,7 +121,24 @@ func NewWithStore(token string, cabinet store.Cabinet) (*State, error) {
...
 	}
 
-	return NewFromSession(s, cabinet), nil
+	return newWithAutoRescale(s, cabinet), nil
+}
+
+func newWithAutoRescale(s *session.Session, cabinet store.Cabinet) *State {
+	state := NewFromSession(s, cabinet)
+	state.ShardManager.Rescale = func() *shard.Manager {
+		state.Reset()
```
diamondburned: `Reset` is not only called here but also on `Ready` events, which would cause the state to be incomplete when the shards are starting up. A way to fix this would probably be to have a `noResetOnReady bool` that prevents the handlers from resetting the state there, so we can do it here.

This call should also be moved to after the `NewManager` call.

state/state.go
```diff
@@ -6,6 +6,8 @@ import (
 	"context"
 	"sync"
 
+	"github.com/diamondburned/arikawa/v2/gateway/shard"
+
```
diamondburned: Random new line?

diamondburned (changes requested):

voice/voice.go
mavolin: I think instead of having two methods, one for gateway and one for shard.Manager, it might be simpler to just have an expose `Intents` constant, that users add themselves.

diamondburned: That works as well.

state/state.go
mavolin: If we expect people to implement their own shard manager, we should probably expose `noResetOnReady`.

PR #195, opened 2021-05-12
interaction: Implement buttons

chanbakjsd2:
This is currently an API in private beta so the PR is not usable unless you are part of it.

The API is subject to change but this PR should reflect a relatively new version of the API.

### DIFF
--- a/discord/component.go
+++ b/discord/component.go
@@ -1,0 +1,93 @@
+package discord
+
+import "encoding/json"
+
+// ComponentType is the type of a component.
+type ComponentType uint
+
+const (
+	ActionRowComponent ComponentType = iota + 1
+	ButtonComponent
+)
+
+// Component is a component that can be attached to an interaction response.
+type Component interface {
+	json.Marshaler
+	Type() ComponentType
+}
+
+// ActionRow is a row of components at the bottom of a message.
+type ActionRow struct {
+	Components []Component `json:"components"`
+}
+
+// Type implements the InteractionComponent interface.
+func (ActionRow) Type() ComponentType {
+	return ActionRowComponent
+}
+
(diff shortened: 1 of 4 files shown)


discord/component.go
```diff
@@ -0,0 +1,93 @@
...
+	SecondaryButton                        // Grey button.
+	SuccessButton                          // Green button.
+	DangerButton                           // Red button.
+	LinkButton                             // Button that navigates to a URL.
+)
+
+// ButtonEmoji is the emoji displayed on the button before the text.
+type ButtonEmoji struct {
+	Name     string `json:"name"`
+	ID       string `json:"id"`
```
diamondburned: Is there a reason this doesn't have a specific type?

diamondburned (changes requested):

discord/component.go
chanbakjsd2: It has been updated to be a `discord.EmojiID`.

diamondburned (dismissed):

diamondburned:
Holding until the official API is stable.

yyewolf:
Aren't buttons also available on normal embeds (not using slash commands) ? (I've seen a person using them like this)

diamondburned:
cc @chanbakjsd

chanbakjsd2:
I'm assuming you mean buttons below embeds?
Buttons are an extra field to the reply of interactions so you can reply with the embed field and the component field to achieve that effect. The example uses `content` simply because it's the easiest (no need to construct a new embed when the point is the big struct of buttons).

matthewpi:
https://github.com/discord/discord-api-docs/pull/3007 was just merged

chanbakjsd2:
Unblocking because buttons API has been officially released.

diamondburned (approved):

PR #196, opened 2021-05-20
discord: Add certified moderator user flag

chanbakjsd2:
A new user flag has been documented in discord/discord-api-docs@55cd3d7.
This commit adds the flag to arikawa.

### DIFF
--- a/discord/user.go
+++ b/discord/user.go
@@ -82,6 +82,7 @@ const (
 	_
 	VerifiedBot
 	VerifiedBotDeveloper
+	CertifiedModerator
 )
 
 type UserNitro uint8


discord/user.go
```diff
@@ -82,6 +82,7 @@ const (
 	_
 	VerifiedBot
 	VerifiedBotDeveloper
+	CertifiedModerator
```
mavolin, line 85: They [changed](https://github.com/discord/discord-api-docs/pull/2946/commits/fc2b436eb25b5b17d7fcd31e8dba16bc3ae385e5) the name to "Discord Certified Developer", so I think we should do the same.

mavolin (changes requested):

discord/user.go
chanbakjsd: Naming it `CertifiedModerator` as opposed to `DiscordCertifiedModerator` is a conscious choice to prevent stuttering (`discord.DiscordCertifiedModerator`) as they are in `package discord`.

This is similar to `discord.Employee` as opposed to `discord.DiscordEmployee`.

mavolin: You're right, `Employee` does the same, therefore resolving this.

mavolin (approved):

PR #197, opened 2021-05-21
discord: Helper methods & doc fixes

twoscott:
Just some methods I thought would be convenient. I haven't used most of these structs yet but I figured the CreatedAt methods should be on any object that has a snowflake directly identifying it (not sure if I missed any).

### DIFF
--- a/discord/emoji.go
+++ b/discord/emoji.go
@@ -36,7 +36,7 @@ type Emoji struct {
 	// This field is only available for custom emojis.
 	Animated bool `json:"animated,omitempty"`
 	// Available specifies whether the emoji can be used.
-	// This may be false tue to loss of Server Boosts.
+	// This may be false due to loss of Server Boosts.
 	//
 	// This field is only available for custom emojis.
 	Available bool `json:"available,omitempty"`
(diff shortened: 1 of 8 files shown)


diamondburned:
Is there a reason why `CreatedAt` returns `discord.Timestamp` and not `time.Time`?

twoscott:
I honestly just wasn't sure which to use, figured you might prefer to use `discord.Timestamp` to keep consistency but I could see `time.Time` working just as well if not better

diamondburned:
`discord.Timestamp` would indeed be more consistent, but since these are methods, it doesn't really matter. The type was made just so JSON can be marshaled properly, so it doesn't matter as far as usage goes.

discord/emoji.go
```diff
@@ -35,12 +36,29 @@ type Emoji struct {
...
 }
 
+// IsCustom returns whether the emoji is a custom emoji.
+func (e Emoji) IsCustom() bool {
+	return !e.IsUnicode()
+}
+
+// IsUnicode returns whether the emoji is a unicode emoji.
+func (e Emoji) IsUnicode() bool {
+	return e.ID.IsNull()
```
diamondburned: It would be better to use `.IsValid()` here.

diamondburned (changes requested):

diamondburned (approved):

PR #213, opened 2021-05-30
v3: Add paginating State.Messages

mavolin:


state/store/defaultstore/message.go
```diff
@@ -73,68 +73,129 @@ func (s *Message) MaxMessages() int {
...
+		return true
+	}
+
+	// Two cases remain, the timestamps are equal or message is older than our
+	// first message. So we compare timestamps. If they are equal, make sure
+	// messages doesn't contain a message with the same id, in order to prevent
+	// insertion of a duplicate.
+	// ID timestamps are used, as they provide millisecond accuracy in contrast
+	// to the second accuracy of discord.Message.Timestamp.
+	ts := message.ID >> 22
```
diamondburned: The timestamp can be obtained with `message.ID.Time()`.

state/store/defaultstore/message.go
```diff
@@ -73,68 +73,129 @@ func (s *Message) MaxMessages() int {
...
-		// right after this message (or before this message in the list; i-1).
-		if message.ID > msgs.messages[insertAt].ID {
-			break
+// shouldPrependMessage checks if the passed message may be prepended to the
+// passed message slice, that is ordered from latest to oldest.
+//
+// shouldPrependMessage is biased as it will return true if the timestamps
+// of the passed message and the latest message match, even though the true
+// order cannot be determined in that case.
+func shouldPrependMessage(message discord.Message, messages []discord.Message) bool {
```
diamondburned: Wouldn't it be more ideal to have an API that returns the index to be inserted in the list? `-1` can be returned to mean ignore.

diamondburned (changes requested):

state/store/defaultstore/message.go
mavolin: Insertion into the slice, as in somewhere in the middle, is never desired though, since the messages stored are guaranteed to be complete. The only reason there is even a check for duplicates, is to prevent toctou problems where in-between the time of checking and calling `MessageSet` there could be another function that already added that message. Gaps, however, are guaranteed not to occur, which is why I don't think we even need indexes.

diamondburned: That's a fair point, though I still think we could return a number to indicate prepending or appending without needing two functions. Something like `1` for append, `-1` for prepend, and `0` for ignore?

state/store/defaultstore/message.go
mavolin: I know, this syntax was just shorter to compare, and (subjectively) more readable: `ts < message[i].ID>>22` vs. `ts.Equal(message[i].ID.Time())`. I can change it though, if you want.

state/store/defaultstore/message.go
mavolin: Sure, I'll get on doing that tomorrow.

state/store/defaultstore/message.go
diamondburned: I think it's better to change it, since `>>22` doesn't seem immediately obvious.

diamondburned:
I'm still somewhat skeptical of this prepend/append-only thing, but I guess it doesn't matter for bot usages, and time will tell as far as user clients go.

diamondburned (approved):

diamondburned:
@mavolin can you fix the merge conflict?

PR #226, opened 2021-06-10
v3: Rework the shard manager

diamondburned:
This pull request reworks the shard manager to be much less intrusive on other packages. Instead of forcing all components to store the shard manager, it now acts as some sort of wrapper over components and manage them directly. The caller no longer needs to do `t.ShardManager.M()` to access the gateway.

A side effect of this pull request is that the `Gateway` type is now embedded into all components. This may make code slightly more confusing, but it shouldn't be too bad.

**Note:** This pull request can NOT be merged from GitHub. It has to be done from the CLI. Reviewers may approve of the pull request still, however. The actual relevant changes are listed at https://github.com/diamondburned/arikawa/pull/226/commits/adf029bd7c4ea687dddadb6c3f1fcdb4956f012b.


_example/sharded/main.go
```diff
@@ -0,0 +1,61 @@
...
+		s.Gateway.AddIntents(gateway.IntentDirectMessages)
+
+		s.AddHandler(func(c *gateway.MessageCreateEvent) {
+			_, shardIx := m.FromGuildID(c.GuildID)
+			log.Println(c.Author.Tag(), "sent", c.Content, "on shard", shardIx)
+		})
+	})
+
+	m, err := shard.NewManager("Bot "+token, newShard)
+
```
mavolin: Remove this blank line.
```suggestion
```

api/bot.go
```diff
@@ -0,0 +1,41 @@
...
+
+// GatewayURL asks Discord for a Websocket URL to the Gateway.
+func GatewayURL() (string, error) {
+	var g BotData
+
+	if err := httputil.NewClient().RequestJSON(&g, "GET", EndpointGateway); err != nil {
+		return "", err
+	}
+
+	return g.URL, nil
```
mavolin: This can be simplified to:

```suggestion
	var g BotData
	return g.URL, httputil.NewClient().RequestJSON(&g, "GET", EndpointGateway)
```

gateway/gateway.go
```diff
@@ -124,22 +97,16 @@ type Gateway struct {
...
-	OnScalingRequired func()
-
 	// AfterClose is called after each close or pause. It is used mainly for
 	// reconnections or any type of connection interruptions.
 	//
 	// Constructors will use a no-op function by default.
 	AfterClose func(err error)
 
-	waitGroup sync.WaitGroup
+	onScalingRequired func()
```
mavolin: There is no `Scaling Required` error code. It's actually `Sharding Required` and I mistyped it. The original shard manager PR fixes this, and so should this.
```suggestion
	onShardingRequired func()
```

gateway/gateway.go
```diff
@@ -318,6 +278,18 @@ func (g *Gateway) SessionID() string {
...
 }
 
+// OnScalingRequired sets the function to be called if Discord closes with error
+// code 4011 aka Scaling Required. When called, the Gateway will already be
+// closed, and can (after increasing the number of shards) be reopened using
+// Open. Reconnect or ReconnectCtx, however, will not be available as the
+// session is invalidated.
+//
+// The gateway will completely halt what it's doing in the background when this
+// callback is called.
```
mavolin:
```suggestion
// OnScalingRequired sets the function to be called if Discord closes with error
// code 4011 aka Sharding Required. When called, the Gateway will already be
// closed, and can (after increasing the number of shards) be reopened using
// Open. Reconnect or ReconnectCtx, however, will not be available as the
// session is invalidated.
//
// The gateway will completely halt what it's doing in the background when this
// callback is called.
```

gateway/gateway.go
```diff
@@ -318,6 +278,18 @@ func (g *Gateway) SessionID() string {
...
+func (g *Gateway) OnScalingRequired(fn func()) {
+	g.onScalingRequired = fn
```
mavolin:
```suggestion
func (g *Gateway) OnScalingRequired(fn func()) {
	g.onScalingRequired = fn
```
```suggestion
func (g *Gateway) OnShardingRequired(fn func()) {
	g.onShardingRequired = fn
```

gateway/gateway.go
```diff
@@ -467,7 +439,7 @@ func (g *Gateway) start(ctx context.Context) error {
 			g.sessionID = ""
 			g.sessionMu.Unlock()
 
-			g.OnScalingRequired()
+			g.onScalingRequired()
```
mavolin:
```suggestion
			g.onShardingRequired()
```

gateway/shard/manager.go
```diff
@@ -0,0 +1,314 @@
...
+// Discord.
+func NewManager(token string, new NewShardFunc) (*Manager, error) {
+	id := gateway.DefaultIdentifier(token)
+
+	url, err := updateIdentifier(context.Background(), id)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get gateway info")
+	}
+
+	return NewIdentifiedManagerWithURL(url, id, new)
```
mavolin: `new` is a built-in function so, even though it's not used in the function, we should opt for another name.
```suggestion
func NewManager(token string, fn NewShardFunc) (*Manager, error) {
	id := gateway.DefaultIdentifier(token)

	url, err := updateIdentifier(context.Background(), id)
	if err != nil {
		return nil, errors.Wrap(err, "failed to get gateway info")
	}

	return NewIdentifiedManagerWithURL(url, id, fn)
```

mavolin (changes requested):

gateway/shard/manager.go
```diff
@@ -0,0 +1,314 @@
...
+		}
+
+		m.shards[i].Shard, err = new(&m, &m.shards[i].id)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create shard %d/%d", i, len(m.shards)-1)
+		}
+	}
+
+	return &m, nil
+}
```
mavolin: `new` should be replaced here as well.
```suggestion
func NewIdentifiedManagerWithURL(
	url string, id *gateway.Identifier, fn NewShardFunc) (*Manager, error) {

	m := Manager{
		gatewayURL: gateway.AddGatewayParams(url),
		shards:     make([]shard, id.Shard.NumShards()),
		new:        fn,
	}

	var err error

	for i := range m.shards {
		data := id.IdentifyData
		data.Shard = &gateway.Shard{i, len(m.shards)}

		m.shards[i] = shard{
			id: gateway.Identifier{
				IdentifyData:        data,
				IdentifyShortLimit:  id.IdentifyShortLimit,
				IdentifyGlobalLimit: id.IdentifyGlobalLimit,
			},
		}

		m.shards[i].Shard, err = fn(&m, &m.shards[i].id)
		if err != nil {
			return nil, errors.Wrapf(err, "failed to create shard %d/%d", i, len(m.shards)-1)
		}
	}

	return &m, nil
}
```

gateway/shard/manager.go
```diff
@@ -0,0 +1,314 @@
...
+	go m.rescale()
+}
+
+func (m *Manager) rescale() {
+	m.mutex.Lock()
+
+	// Exit if we're already rescaling.
+	if m.rescaling != nil {
+		return
+	}
```
mavolin: The mutex won't be unlocked, when returning in L224
```suggestion
	m.mutex.Lock()

	// Exit if we're already rescaling.
	if m.rescaling != nil {
		m.mutex.Unlock()
		return
	}
```

diamondburned:
Merged into v3 branch manually.

PR #230, opened 2021-06-17
v3: Update more endpoints to Discord API v9

samhza:
Update message editing/sending for both API and webhooks.

### DIFF
--- a/api/message.go
+++ b/api/message.go
@@ -1,12 +1,16 @@
 package api
 
 import (
+	"mime/multipart"
+	"strconv"
+
 	"github.com/pkg/errors"
 
 	"github.com/diamondburned/arikawa/v3/discord"
 	"github.com/diamondburned/arikawa/v3/internal/intmath"
 	"github.com/diamondburned/arikawa/v3/utils/httputil"
 	"github.com/diamondburned/arikawa/v3/utils/json/option"
+	"github.com/diamondburned/arikawa/v3/utils/sendpart"
 )
 
 const (
@@ -220,17 +224,17 @@ func (c *Client) SendTextReply(
 	})
 }
 
-// SendEmbed posts an Embed to a guild text or DM channel.
+// SendEmbeds sends embeds to a guild text or DM channel.
 //
 // If operating on a guild channel, this endpoint requires the SEND_MESSAGES
 // permission to be present on the current user.
 //
 // Fires a Message Create Gateway event.
-func (c *Client) SendEmbed(
-	channelID discord.ChannelID, e discord.Embed) (*discord.Message, error) {
+func (c *Client) SendEmbeds(
+	channelID discord.ChannelID, e ...discord.Embed) (*discord.Message, error) {
 
 	return c.SendMessageComplex(channelID, SendMessageData{
-		Embed: &e,
+		Embeds: e,
 	})
 }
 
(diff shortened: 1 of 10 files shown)


samhza:
The `embed` field for editing/sending messages has been deprecated in favor of an `embeds` field, allowing for sending/editing multiple embeds in one message. Should EditMessage/EditEmbed's API be kept the same but internally using `Embeds: &[]discord.Embed{embed}` instead of `Embed: embed`? Either way, I think we should remove the `Embed` field from {Send,Edit}MessageData.

https://discord.com/developers/docs/resources/channel#create-message

diamondburned:
@samhza 

I agree that we should remove the singular Embed field and use the multiple one instead.

api/message.go
```diff
@@ -230,7 +234,7 @@ func (c *Client) SendEmbed(
 	channelID discord.ChannelID, e discord.Embed) (*discord.Message, error) {
```
diamondburned: Would `SendEmbed(discord.ChannelID, ...discord.Embed)` be a better API for this?

api/message.go
```diff
@@ -259,11 +263,13 @@ func (c *Client) SendEmbedReply(
 // Fires a Message Create Gateway event.
 func (c *Client) SendMessage(
 	channelID discord.ChannelID, content string, embed *discord.Embed) (*discord.Message, error) {
```
diamondburned: I think this can also do `SendMessage(discord.ChannelID, string, ...discord.Embed)`.

api/message.go
```diff
@@ -318,7 +339,7 @@ func (c *Client) EditEmbed(
 	channelID discord.ChannelID, messageID discord.MessageID, embed discord.Embed) (*discord.Message, error) {
```
diamondburned: (above 2 comments)

utils/sendpart/sendpart.go
```diff
@@ -37,16 +38,15 @@ type DataMultipartWriter interface {
 	httputil.MultipartWriter
 }
 
-// POST sends a POST request using client to the given URL and unmarshal the
-// body into v if it's not nil. It will only send using multipart if files is
-// true.
-func POST(c *httputil.Client, data DataMultipartWriter, v interface{}, url string) error {
+// Do sends an HTTP request using client to the given URL and unmarshals the
+// body into v if it's not nil. It will only send using multipart if needed.
+func Do(method string, c *httputil.Client, data DataMultipartWriter, v interface{}, url string) error {
```
diamondburned: The order of these arguments should be `Do(c *httputil.Client, method string, data DataMultipartWriter, v interface{}, url string) error`.

utils/sendpart/sendpart.go
```diff
@@ -61,6 +61,20 @@ func POST(c *httputil.Client, data DataMultipartWriter, v interface{}, url strin
 	return json.DecodeStream(body, v)
 }
 
+// PATCH sends a PATCH request using client to the given URL and unmarshals the
+// body into v if it's not nil. It will only send using multipart if needed.
+// It is equivalent to calling Do with http.MethodPost
+func POST(c *httputil.Client, data DataMultipartWriter, v interface{}, url string) error {
+	return Do(http.MethodPost, c, data, v, url)
```
diamondburned: The `http.Method*` constants probably shouldn't be used here to avoid importing `net/http` just for those constants.

diamondburned (changes requested):

diamondburned (approved):

PR #232, opened 2021-06-18
make OverboundError.Error() a pointer receiver

samhza:


diamondburned:
Would it be better to make `OverboundError` have pointer receivers instead?

samhza:
I don't know which option would be more idiomatic

diamondburned:
Probably the pointer receiver one.

diamondburned (approved):

PR #240, opened 2021-07-04
Fix #238 branch for region/voice server changes

### LINKED ISSUE #238: Voice: v2: ReadPacket() never finishes when region changes
When the voice server's region changes, `ReadPacket()` will never finish as it (presumably) waits indefinitely to read a packet from the old UDP voice server.

I believe the solution here is to `SetReadDeadline` to something like 1-2 seconds on the underlying connection, read, check if the gateway has changed, otherwise re-attempt the read and loop.

1lann:
I tested that this works, region changes are handled gracefully with packets from `ReadPacket` resuming like nothing happened.

### DIFF
--- a/voice/udp/manager.go
+++ b/voice/udp/manager.go
@@ -3,8 +3,6 @@ package udp
 import (
 	"context"
 	"net"
-	"runtime/debug"
-	"strings"
 	"sync"
 	"time"
 
@@ -170,7 +168,6 @@ func (m *Manager) ReadPacket() (p *Packet, err error) {
 	p, err = conn.ReadPacket()
 	if err != nil && conn.IsClosed() {
 		wsutil.WSDebug("UDP connection was closed, re-attempting read...")
-		// UDP connection was closed during read, re-attempt the read.
 		return m.ReadPacket()
 	}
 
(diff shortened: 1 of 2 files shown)


voice/udp/manager.go
```diff
@@ -160,7 +165,13 @@ func (m *Manager) ReadPacket() (p *Packet, err error) {
 		return nil, ErrManagerClosed
 	}
 
-	return conn.ReadPacket()
+	p, err = conn.ReadPacket()
+	if err != nil && conn.IsClosed() {
+		wsutil.WSDebug("UDP connection was closed, re-attempting read...")
+		return m.ReadPacket()
```
diamondburned: I think it's a better idea to make this a loop. Using recursion for this seems overkill.

voice/udp/manager.go
```diff
@@ -177,14 +188,9 @@ func (m *Manager) Write(b []byte) (n int, err error) {
 // acquireConn acquires the current connection and releases the lock, returning
 // the connection at that point in time. Nil is returned if Manager is closed.
 func (m *Manager) acquireConn() *Connection {
-	m.stopMu.Lock()
-	defer m.stopMu.Unlock()
-
 	select {
 	case m.connLock <- struct{}{}:
```
diamondburned: This should probably be changed to just

```go
m.connLock <- struct{}{}
defer func() { <-m.connLock }()
```

voice/udp/manager.go
```diff
@@ -127,6 +128,10 @@ func (m *Manager) PauseAndDial(ctx context.Context, addr string, ssrc uint32) (*
 		conn.ResetFrequency(m.frequency, m.timeIncr)
 	}
 
+	if m.conn != nil {
```
diamondburned: Is there a reason why this is done after connecting and not before? It seems to me that `connLock` is acquired anyway, so there's no reason to not close it before.

diamondburned (changes requested):

voice/udp/manager.go
1lann: tfw no tail call optimizatoin, sure.

voice/udp/manager.go
1lann: ¯\\\_(ツ)\_/¯ sounds good to me

diamondburned (approved):

PR #241, opened 2021-07-04
v3: voice: udp.Manager for seamless reconnection

### LINKED ISSUE #238: Voice: v2: ReadPacket() never finishes when region changes

diamondburned:
This pull request breaks the previous v2 voice API to add in a `udp.Manager` that effectively manages reconnection of the voice UDP connection. This fixes a previous issue where changing the voice region would break the voice connection.

The biggest breaking change in this PR (as far as voice sending goes) actually has nothing to do with this fix. The change is written below:

```diff
 func (s *Session) JoinChannelCtx(
-	ctx context.Context, gID discord.GuildID, cID discord.ChannelID, mute, deaf bool) error {
+	ctx context.Context, chID discord.ChannelID, mute, deaf bool) error {
```

This change is made primarily for convenience.

A small side effect of this change is that the minimum supported Go version has been bumped to 1.16 for some reason. By the time v3 is finalized, however, Go 1.17 would've already been released, so this is fine.

This pull request fixes #238.

### DIFF
--- a/api/channel.go
+++ b/api/channel.go
@@ -175,6 +175,11 @@ type ModifyChannelData struct {
 	//
 	// Channel Types: Voice
 	VoiceUserLimit option.NullableUint `json:"user_limit,omitempty"`
+	// RTCRegionID is the channel voice region id. It will be determined
+	// automatically set, if omitted.
+	//
+	// Channel Types: Voice
+	RTCRegionID option.NullableString `json:"rtc_region,omitempty"`
 	// Overwrites are the channel or category-specific permissions.
 	//
 	// Channel Types: Text, News, Store, Voice, Category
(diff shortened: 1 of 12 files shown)


1lann:
I've been using the 238 branch for a while (only like the last 3 weeks cause I totally forgot to switch to that branch for the project I was working on lol) for a persistent bot that stays in voice chat and it's been working great, it doesn't seem to noticably permanently lose connection from voice chat anymore. I've also had no problems with it (apart from #242, I haven't been running it for weeks with `-race` however). I think it might be good enough™️ to merge anyhow, so it makes it into v3.

diamondburned:
Interesting. I'll rebase it tomorrow and see how feasible it is. Chances are more testing would have to be done to ensure that the gateway refactor works with the PR.

diamondburned:
A few people have tested this branch, and it appears to be working normally, except for the case when the user moves the bot around.

PR #243, opened 2021-07-10
v3: discord: raise embed description limit to 4096 characters

starshine-sys:
Discord recently raised the character limit for embed descriptions to 4096 characters (https://discord.com/developers/docs/resources/channel#embed-limits), but this isn't reflected in embed.Validate() yet.


diamondburned:
Is this only for v9 API?

starshine-sys:
I'm not sure if it's just for v9, I've only tested it with v9 though.

PR #267, opened 2021-08-17
v3: discord: add banner & accent_color field to User struct

ghost:
Reference: https://github.com/discord/discord-api-docs/pull/3448

### DIFF
--- a/discord/user.go
+++ b/discord/user.go
@@ -25,6 +25,9 @@ type User struct {
 
 	Locale string `json:"locale,omitempty"`
 	Email  string `json:"email,omitempty"`
+
+	Banner      Hash `json:"banner,omitempty"`
+	AccentColor int  `json:"accent_color,omitempty"`
 }
 
 // CreatedAt returns a time object representing when the user was created.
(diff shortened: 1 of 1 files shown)


discord/user.go
```diff
@@ -25,6 +25,9 @@ type User struct {
 
 	Locale string `json:"locale,omitempty"`
 	Email  string `json:"email,omitempty"`
+
+	Banner      Hash `json:"banner,omitempty"`
+	AccentColor int  `json:"accent_color,omitempty"`
```
diamondburned: This field should be of type `Color`.

diamondburned (changes requested):

discord/user.go
ghost: :+1: Fixed

discord/user.go
```diff
@@ -25,6 +25,9 @@ type User struct {
 
 	Locale string `json:"locale,omitempty"`
 	Email  string `json:"email,omitempty"`
+
+	Banner      Hash  `json:"banner,omitempty"`
+	AccentColor Color `json:"accent_color,omitempty"`
```
diamondburned: On second thought, this field could probably be called `Accent`, since it's
shorter but is just as descriptive with the type name.

diamondburned (changes requested):

discord/user.go
ghost: :+1: Updated

diamondburned (approved):

discord/user.go
```diff
@@ -71,6 +74,22 @@ func (u User) AvatarURLWithType(t ImageType) string {
...
+}
+
+// BannerURLWithType returns the URL of the Banner Image using the passed type.
+// If the user has no Banner, an empty string will be returned.
+func (u User) BannerURLWithType(t ImageType) string {
+	if u.Banner == "" {
+		return ""
+	}
+
+	return "https://cdn.discordapp.com/banners/" + u.ID.String() + "/" + t.format(u.Banner)
```
diamondburned, line 90: This should probably be changed to the `api.Endpoint*` variables in the future.

PR #270, opened 2021-08-19
v3: api: *Client.ActiveThreads returns ActiveThread struct

ghost:
image

### DIFF
--- a/api/channel.go
+++ b/api/channel.go
@@ -482,8 +482,8 @@ type ActiveThread struct {
 
 // ActiveThreads returns all the active threads in the guild, including public
 // and private threads.
-func (c *Client) ActiveThreads(guildID discord.GuildID) ([]ActiveThread, error) {
-	var t []ActiveThread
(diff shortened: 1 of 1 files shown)


api/channel.go
```diff
@@ -482,8 +482,8 @@ type ActiveThread struct {
 
 // ActiveThreads returns all the active threads in the guild, including public
 // and private threads.
-func (c *Client) ActiveThreads(guildID discord.GuildID) ([]ActiveThread, error) {
-	var t []ActiveThread
+func (c *Client) ActiveThreads(guildID discord.GuildID) (ActiveThread, error) {
```
diamondburned: It's probably best that `type ActiveThread` be named `ActiveThreads` instead,
and the method should return `(*ActiveThreads, error)` instead.

diamondburned (changes requested):

api/channel.go
ghost: Updated

PR #271, opened 2021-08-24
api: Store embed produced by discord.Embed.Validate in sent embeds

mavolin:


diamondburned (approved):

diamondburned:
I find it weird that validation would mutate the embeds in the first place. This
might lead to some unexpected cases where the user has their data unexpectedly
modified.

PR #273, opened 2021-08-30
v3: Discord: change type of SelectComponent.MinValues to option.Int

mavolin:
This PR changes the type of `discord.SelectComponent.MinValues` to `option.Int` to allow `0` to be set as a value, as allowed per [the official docs](https://discord.com/developers/docs/interactions/message-components#select-menu-object-select-menu-structure).
Because simply importing package `option` into package `discord` would cause an import cycle, as package `option` also imports package `discord`, I moved all option types that use `discord` types into package `discord`. In the case of `option.NullableColor` I simply refactored `discord.Color` to a signed `int`, and used negative numbers to signal nullability. `option.Color` could just be deleted, as it was unused, and only `option.Seconds` actually required a separate type, `discord.OptionalSeconds`.


diamondburned:
This is quite unfortunate. Perhaps package `option` should entirely be in
`discord`.

diamondburned (dismissed):

diamondburned (approved):

PR #274, opened 2021-09-05
*: move Interaction to discord, update CommandCreateData

samhza:
- Moved gateway.InteractionCreateData to discord.Interaction, now
gateway.InteractionCreateData is a struct that wraps
discord.Interaction.
- Split InteractionData into CommandInteractionData and
ComponentInteractionData.
- Renamed ButtonInteraction to ComponentInteraction.
- Updated api.CommandCreateData to add new fields.

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -1,6 +1,8 @@
 package api
 
 import (
+	"encoding/json"
+
 	"github.com/diamondburned/arikawa/v3/discord"
 	"github.com/diamondburned/arikawa/v3/utils/httputil"
 )
@@ -9,9 +11,49 @@ var EndpointApplications = Endpoint + "applications/"
 
 // https://discord.com/developers/docs/interactions/slash-commands#create-global-application-command-json-params
 type CreateCommandData struct {
-	Name        string                  `json:"name"`
-	Description string                  `json:"description"`
-	Options     []discord.CommandOption `json:"options"`
+	Name                string                  `json:"name"`
+	Description         string                  `json:"description"`
+	Options             []discord.CommandOption `json:"options,omitempty"`
+	NoDefaultPermission bool                    `json:"-"`
+	Type                discord.CommandType     `json:"type,omitempty"`
+}
+
+func (c CreateCommandData) MarshalJSON() ([]byte, error) {
+	type RawCreateCommandData CreateCommandData
+	cmd := struct {
+		RawCreateCommandData
+		DefaultPermission bool `json:"default_permission"`
+	}{RawCreateCommandData: (RawCreateCommandData)(c)}
+
+	// Discord defaults default_permission to true, so we need to invert the
+	// meaning of the field (>No<DefaultPermission) to match Go's default
+	// value, false.
+	cmd.DefaultPermission = !c.NoDefaultPermission
+
+	return json.Marshal(cmd)
+}
+
+func (c *CreateCommandData) UnmarshalJSON(data []byte) error {
+	type RawCreateCommandData CreateCommandData
+	cmd := struct {
+		*RawCreateCommandData
+		DefaultPermission bool `json:"default_permission"`
+	}{RawCreateCommandData: (*RawCreateCommandData)(c)}
+	if err := json.Unmarshal(data, &cmd); err != nil {
+		return err
+	}
+
+	// Discord defaults default_permission to true, so we need to invert the
+	// meaning of the field (>No<DefaultPermission) to match Go's default
+	// value, false.
+	c.NoDefaultPermission = !cmd.DefaultPermission
+
+	// Discord defaults type to 1 if omitted.
+	if c.Type == 0 {
+		c.Type = discord.ChatInputCommand
+	}
+
+	return nil
 }
 
 func (c *Client) Commands(appID discord.AppID) ([]discord.Command, error) {
--- a/discord/application.go
+++ b/discord/application.go
@@ -13,6 +13,8 @@ import (
 type Command struct {
 	// ID is the unique id of the command.
 	ID CommandID `json:"id"`
+	// Type is the type of command.
+	Type CommandType `json:"type,omitempty"`
 	// AppID is the unique id of the parent application.
 	AppID AppID `json:"application_id"`
 	// GuildID is the guild id of the command, if not global.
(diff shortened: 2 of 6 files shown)


api/application.go
```diff
@@ -9,9 +11,36 @@ var EndpointApplications = Endpoint + "applications/"
 
 // https://discord.com/developers/docs/interactions/slash-commands#create-global-application-command-json-params
 type CreateCommandData struct {
-	Name        string                  `json:"name"`
-	Description string                  `json:"description"`
-	Options     []discord.CommandOption `json:"options"`
+	Name                string                  `json:"name"`
+	Description         string                  `json:"description"`
+	Options             []discord.CommandOption `json:"options"`
+	NoDefaultPermission bool                    `json:"default_permission"`
```
diamondburned: This should have a `json:"-"` tag so that it doesn't get confused as a valid
JSON field (because it isn't).

The field could be overridden inside a JSON object wrapper:

```go
type rawCreateCommandData CreateCommandData

type createCommandData struct {
	rawCreateCommandData
	DefaultPermission bool `json:"default_permission"`
}
```

discord/application.go
```diff
@@ -13,6 +13,8 @@ import (
 type Command struct {
 	// ID is the unique id of the command.
 	ID CommandID `json:"id"`
+	// Type is the type of command and defaults to ChatInputCommand if unset.
+	Type CommandType `json:"type,omitempty"`
```
diamondburned, line 17: Since `Type` has a default, the default should probably be set in Command's
`MarshalJSON` as well, so the user doesn't have to check `Type == 0 || Type ==
ChatInputCommand`.

discord/interaction.go
```diff
@@ -0,0 +1,129 @@
...
+// https://discord.com/developers/docs/topics/gateway#interactions
+type Interaction struct {
+	ID        InteractionID   `json:"id"`
+	AppID     AppID           `json:"application_id"`
+	Type      InteractionType `json:"type"`
+	Data      InteractionData `json:"data,omitempty"`
+	ChannelID ChannelID       `json:"channel_id,omitempty"`
+	Token     string          `json:"token"`
+	Version   int             `json:"version"`
+	Message   *Message        `json:"message"`
```
diamondburned: Message is optional; the tag should indicate that: `json:"message,omitempty"`.

A comment should also be added to clarify when this field is present, or at
least to clarify that it is optional.

discord/interaction.go
```diff
@@ -0,0 +1,129 @@
...
+	return ComponentInteraction
+}
+
+type CommandInteractionData struct {
+	ID      CommandID           `json:"id"`
+	Name    string              `json:"name"`
+	Options []InteractionOption `json:"options"`
+}
+
+func (c CommandInteractionData) Type() InteractionType {
```
diamondburned: Change the receiver to take a `*CommandInteractionData`.

discord/interaction.go
```diff
@@ -0,0 +1,129 @@
...
+	return CommandInteraction
+}
+
+type UnknownInteractionData struct {
+	json.Raw
+	typ InteractionType
+}
+
+func (u UnknownInteractionData) Type() InteractionType {
```
diamondburned: See above comment.

discord/interaction.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+type InteractionOption struct {
+	Name    string              `json:"name"`
+	Value   json.Raw            `json:"value"`
+	Options []InteractionOption `json:"options"`
+}
+
+func (o InteractionOption) String() string {
+	val := bytes.Trim([]byte(o.Value), `"`)
```
diamondburned: Use `strconv.Unquote` first, then `Trim` as a fallback. Just using `Trim` alone
will not work properly on a string.

discord/interaction.go
```diff
@@ -0,0 +1,129 @@
...
+	return u.typ
+}
+
+type InteractionOption struct {
+	Name    string              `json:"name"`
+	Value   json.Raw            `json:"value"`
+	Options []InteractionOption `json:"options"`
+}
+
+func (o InteractionOption) String() string {
```
diamondburned, line 107: Clarify that `String()` will never fail and will always print `Value` as either
a valid string or a raw value of other type.

discord/interaction.go
```diff
@@ -0,0 +1,129 @@
...
+const (
+	PingInteraction InteractionType = iota + 1
+	CommandInteraction
+	ComponentInteraction
+)
+
+// InteractionData holds the data of an interaction.
+// Type assertions should be made on InteractionData to
+// access the underlying data.
+type InteractionData interface {
```
diamondburned, line 66: Clarify that the underlying types should always be pointers.

diamondburned (changes requested):

diamondburned (dismissed):

api/application.go
```diff
@@ -9,9 +11,49 @@ var EndpointApplications = Endpoint + "applications/"
 
 // https://discord.com/developers/docs/interactions/slash-commands#create-global-application-command-json-params
 type CreateCommandData struct {
-	Name        string                  `json:"name"`
-	Description string                  `json:"description"`
-	Options     []discord.CommandOption `json:"options"`
+	Name                string                  `json:"name"`
+	Description         string                  `json:"description"`
+	Options             []discord.CommandOption `json:"options"`
```
mavolin:
```suggestion
	Options             []discord.CommandOption `json:"options,omitempty"`
```

This field is optional, so this field should have `,omitempty` in the tag.

api/application.go
```diff
@@ -9,9 +11,49 @@ var EndpointApplications = Endpoint + "applications/"
...
+		return err
+	}
+
+	// Discord defaults default_permission to true, so we need to invert the
+	// meaning of the field (>No<DefaultPermission) to match Go's default
+	// value, false.
+	c.NoDefaultPermission = !cmd.DefaultPermission
+
+	// Discord defaults type to 1 if omitted.
+	if c.Type == discord.CommandType(0) {
```
mavolin:
```suggestion
	if c.Type == 0 {
```

The explicit conversion of `0` to `discord.CommandType` is unnecessary, as Go can automatically infer the literal's type from `c.Type`.

discord/application.go
```diff
@@ -29,30 +31,51 @@ type Command struct {
 	Options []CommandOption `json:"options,omitempty"`
```
mavolin, line 33:
```suggestion
	//
	// It is only present on ChatInputCommands.
	Options []CommandOption `json:"options,omitempty"`
```
Since `options` is now only valid for `ChatInputCommand`s, that should probably be noted in `Options` doc.

discord/application.go
```diff
@@ -29,30 +31,51 @@ type Command struct {
...
 	}
 
 	// Discord defaults default_permission to true, so we need to invert the
 	// meaning of the field (>No<DefaultPermission) to match Go's default
 	// value, false.
-	c.NoDefaultPermission = !c.NoDefaultPermission
+	c.NoDefaultPermission = !cmd.DefaultPermission
+
+	// Discord defaults type to 1 if omitted.
+	if c.Type == CommandType(0) {
```
mavolin:
```suggestion
	if c.Type == 0 {
```
Again, unnecessary type conversion.

mavolin (changes requested):

mavolin (approved):

PR #275, opened 2021-09-08
gateway: make IdentifyData's intents field an optional.Uint

samhza:
<https://discord.com/developers/docs/topics/gateway#identify>
The `intents` field is required, and 0 is a valid value for intents. Currently if you try to connect with `intents` = 0, you get an error from Discord because the Identify would be sent without an `intents` field.

### DIFF
--- a/bot/extras/middlewares/middlewares_test.go
+++ b/bot/extras/middlewares/middlewares_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/diamondburned/arikawa/v3/session"
 	"github.com/diamondburned/arikawa/v3/state"
 	"github.com/diamondburned/arikawa/v3/state/store"
+	"github.com/diamondburned/arikawa/v3/utils/json/option"
 )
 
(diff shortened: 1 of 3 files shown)


diamondburned:
This breaks if the client connecting is a user account.

Using an optional int would be better.

samhza:
So should I make the `Intents` field an option.Uint? What should be done about `type Intents`

diamondburned:
> So should I make the `Intents` field an option.Uint?

That could work. The user could use the `AddIntents()` method that initializes the field if it's not already. `AddIntents(0)` would do what you want.

samhza:
made the requested changes

diamondburned:
On second thought, it would be a better idea to have a `type Intents
option.Uint` instead of changing the field's type and not using the previous
type anymore.

Also, the CI is failing.

samhza:
I've fixed the tests, CI is now passing

diamondburned (approved):

PR #278, opened 2021-09-28
v3: discord, api, gateway: add per-guild avatars

starshine-sys:
Reference: discord/discord-api-docs#3081

(Had to break the pattern established by User.AvatarURL() and Guild.IconURL() etc, as the member object doesn't include a guild ID)

### DIFF
--- a/api/user.go
+++ b/api/user.go
@@ -46,10 +46,10 @@ func (c *Client) ModifyMe(data ModifySelfData) (*discord.User, error) {
 	)
 }
 
-// ChangeOwnNickname modifies the nickname of the current user in a guild.
+// ModifyCurrentMember modifies the nickname of the current user in a guild.
 //
 // Fires a Guild Member Update Gateway event.
-func (c *Client) ChangeOwnNickname(
+func (c *Client) ModifyCurrentMember(
 	guildID discord.GuildID, nick string) error {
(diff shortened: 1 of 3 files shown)


mavolin (changes requested):
This looks good. However, could you rename `ModifyMe` to `ModifyCurrentUser` to make the distinction between the member and the user method clear.

api/user.go
```diff
@@ -36,8 +36,8 @@ type ModifySelfData struct {
 	AuditLogReason `json:"-"`
 }
 
-// ModifyMe modifies the requester's user account settings.
-func (c *Client) ModifyMe(data ModifySelfData) (*discord.User, error) {
+// ModifyCurrentUser modifies the requester's user account settings.
+func (c *Client) ModifyCurrentUser(data ModifySelfData) (*discord.User, error) {
```
mavolin: `ModifySelfData` should also be renamed to `ModifyCurrentUserData`.

mavolin (changes requested):

api/user.go
starshine-sys: woops, forgot about that!

mavolin (approved):

diamondburned (approved):

PR #279, opened 2021-09-28
v3: discord: add ChannelType field to CommandOption

starshine-sys:
https://github.com/discord/discord-api-docs/pull/3825

Bit hacky because of the `[]uint16`, but it works

### DIFF
--- a/discord/application.go
+++ b/discord/application.go
@@ -97,6 +97,43 @@ type CommandOption struct {
 	Required    bool                  `json:"required"`
 	Choices     []CommandOptionChoice `json:"choices,omitempty"`
 	Options     []CommandOption       `json:"options,omitempty"`
+
+	// If this option is a channel type, the channels shown will be restricted to these types
+	ChannelTypes []ChannelType `json:"-"`
+}
+
+func (c CommandOption) MarshalJSON() ([]byte, error) {
+	type RawOption CommandOption
+	option := struct {
+		RawOption
+		ChannelTypes []uint16 `json:"channel_types,omitempty"`
+	}{RawOption: RawOption(c)}
+
+	// []uint8 is marshalled as a base64 string, so we marshal a []uint64 instead.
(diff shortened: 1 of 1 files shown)


discord/application.go
```diff
@@ -97,6 +97,43 @@ type CommandOption struct {
...
+func (c CommandOption) MarshalJSON() ([]byte, error) {
+	type RawOption CommandOption
+	option := struct {
+		RawOption
+		ChannelTypes []uint16 `json:"channel_types,omitempty"`
+	}{RawOption: RawOption(c)}
+
+	// []uint8 is marshalled as a base64 string, so we marshal a []uint64 instead.
+	if len(c.ChannelTypes) > 0 {
+		for _, t := range c.ChannelTypes {
```
diamondburned, line 115: It's better to preallocate the `option.ChannelTypes` slice here:

```suggestion
		option.ChannelTypes = make([]uint16, 0, len(c.ChannelTypes))
		for _, t := range c.ChannelTypes {
```

discord/application.go
```diff
@@ -97,6 +97,43 @@ type CommandOption struct {
...
+	type RawOption CommandOption
+	cmd := struct {
+		*RawOption
+		ChannelTypes []uint16 `json:"channel_types,omitempty"`
+	}{RawOption: (*RawOption)(c)}
+	if err := json.Unmarshal(data, &cmd); err != nil {
+		return err
+	}
+
+	for _, t := range cmd.ChannelTypes {
```
diamondburned, line 134: Also better to preallocate here:

```suggestion 
	c.ChannelTypes = make([]ChannelType, 0, len(cmd.ChannelTypes))
	for _, t := range cmd.ChannelTypes {
```

Note that we shouldn't have a length check here, because we want to override
`c.ChannelTypes` if it's actually in the JSON.

diamondburned (changes requested):

diamondburned (approved):

PR #280, opened 2021-10-01
Implement autocomplete components

hhhapz:

PR #284, opened 2021-10-17
v3: discord: Application Commands refactor

diamondburned:
### Preamble

This refactor is the first big step in changing my primary goal for Arikawa:
instead of being just another low-level wrapper that has high-level
abstractions, Arikawa should strive to be a library that is high-level and
pleasant to use right from its roots. This shouldn't mean it should encapsulate
so much as to lose its flexibility, but it shouldn't try to keep things 1-to-1
with Discord, either.

### Goal

The goal of this refactor is to make handling application commands as easy as
possible. Prior to this PR, it would be very easy to make a mistake while
programming them, which would cause Discord to yield awfully ambiguous errors.

With this new refactor, most of those mistakes wouldn't even be possible to
make: lots of type safety has been added into the data structures so that the
user has to do less work.

For example, before this PR, the user can easily make nested
`ActionRowComponent`s. This would cause Discord to error out. Now, if the user
tries to do so, the compiler won't even build the code.

Many "union-interface" types now also have their respective constructor
functions to hint the supported types to the user in the generated
documentation. This should make handling those types much clearer.

**Documentation Link**: https://pkg.go.dev/github.com/diamondburned/arikawa/v3@v3.0.0-rc.2.0.20211017004259-b1308dc0080e/discord

---

**Warning**: this PR is currently a draft, and there may be
marshaling/unmarshaling bugs. Please test and report any bugs before approving.
**DO NOT MERGE.**


diamondburned:
There's currently a bug in `_example/buttons`: clicking a button does not echo
any events back. I can't figure out why this is the case.

diamondburned:
Created branch [feat-autocomplete-refactor][b] to track PR #280.

[b]: https://github.com/diamondburned/arikawa/tree/feat-autocomplete-refactor

diamondburned:
Superseded by #289.

PR #285, opened 2021-10-19
Gateway: Fix Context overwrite in Gateway.Open

mavolin:
This PR fixes a bug where the `context.Context` given to `Gateway.Open` would always get overwritten with another context.
Furthermore, it removes a deadline maximum from `wsutil.Websocket.Dial`, which would limit the deadline of the context given to `Dial` to `Websocket.Timeout`, regardless of the deadline set by the caller.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -14,13 +14,14 @@ import (
 	"sync"
 	"time"
 
+	"github.com/gorilla/websocket"
+	"github.com/pkg/errors"
+
 	"github.com/diamondburned/arikawa/v3/api"
 	"github.com/diamondburned/arikawa/v3/internal/moreatomic"
 	"github.com/diamondburned/arikawa/v3/utils/json"
 	"github.com/diamondburned/arikawa/v3/utils/json/option"
 	"github.com/diamondburned/arikawa/v3/utils/wsutil"
-	"github.com/gorilla/websocket"
-	"github.com/pkg/errors"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


diamondburned:
Is it a good idea to not have a timeout at all for connecting? I don't think it's common for people to know to use a timeout here.

mavolin:
`Dial` is only ever called by `gateway.Gateway.Open` and `voicegateway.Gateway.OpenCtx`, both of which accept a context, so forgetting a timeout seems unlikely to me.

mavolin:
If there really is a need for `Dial` to always time out, i.e. never permitting `context.Background()`, the best way to solve this would be to check if there is a deadline, and if not wrap the context with a timeout.
```go
if _, ok := ctx.Deadline(); !ok {
    ctx, cancel = context.WithTimeout(ctx, w.Timeout)
}
```
Still, Imho, `Open` always taking in a context makes this negligible, since this makes very clear that timeouts are advised.

diamondburned:
> Still, Imho, `Open` always taking in a context makes this negligible, since this makes very clear that timeouts are advised.

Not at all. Taking in a context may just be indicative that cancellation is
available, not necessarily that the caller must always use a timeout. Most
people would typically just create a context in `main` and pass it down to every
call, even if the context provides cancellation through SIGINT, meaning that the
same context used for, e.g. event loops, would be used for opening the initial
connection, which means there wouldn't be a timeout.

> the best way to solve this would be to check if there is a deadline, and if not wrap the context with a timeout.

This could work.

diamondburned (approved):

PR #302, opened 2022-01-15
state: fix updating reaction count in reaction remove event

Juby210:


diamondburned:
What case does this fix cover? You should add a comment documenting such cases.

Juby210:
> What case does this fix cover? You should add a comment documenting such cases.

You are incrementing reaction count in add event, but you don't decrement it in remove event.
I have starboard bot and you could just increment star count by deleting reaction and adding it again.

Current code only decrements count if bot user removes reaction: 
image

diamondburned:
Wouldn't that mean the *actual* fix would be to omit the `r.Me` branch completely and keep the `default:` branch?

The `r.Me` check can just be its own `if` outside the `switch case`.

diamondburned (approved):

PR #310, opened 2022-02-09
Modal interaction support

itslychee:
This PR aims to provide support for the recently [released modal interactions](https://canary.discord.com/channels/613425648685547541/697138785317814292/940761396809973790). Additionally, I have added the [text field component](https://discord.com/developers/docs/interactions/message-components#text-inputs) as it's specific to modals, and yeah let me know if i missed anything

### DIFF
--- a/api/interaction.go
+++ b/api/interaction.go
@@ -25,7 +25,7 @@ const (
 	DeferredMessageUpdate
 	UpdateMessage
 	AutocompleteResult
-	ModalInteraction
+	Modal
 )
 
 // InteractionResponseFlags implements flags for an
--- a/api/interaction.go
+++ b/api/interaction.go
@@ -25,7 +25,7 @@ const (
 	DeferredMessageUpdate
 	UpdateMessage
 	AutocompleteResult
-	Modal
+	ModalResponse
 )
 
 // InteractionResponseFlags implements flags for an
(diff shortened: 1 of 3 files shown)


api/interaction.go
```diff
@@ -25,6 +25,7 @@ const (
 	DeferredMessageUpdate
 	UpdateMessage
 	AutocompleteResult
+	Modal
```
diamondburned: Can this be named `ModalResponse`? `Modal` is a bit too ambiguous.

(As a side note, I think it's worth renaming these constants to have the `ResponseType` suffix, so `ModalResponseType`. I'll consider it after this PR.)

api/interaction.go
```diff
@@ -78,6 +79,11 @@ type InteractionResponseData struct {
 	//
 	// During all other events, this should not be provided.
 	Choices *[]AutocompleteChoice `json:"choices"`
+	
+	// CustomID used with the modal
+	CustomID option.NullableString `json:"custom_id,omitempty"`
+	// Title is the heading of the modal window
+	Title    option.NullableString `json:"title,omitempty"`
```
diamondburned: This line isn't formatted. Can you do that?

discord/interaction.go
```diff
@@ -365,6 +368,19 @@ func (o CommandInteractionOption) FloatValue() (float64, error) {
 	return f, err
 }
 
+// ModalInteraction is the submitted modal form
+type ModalInteraction struct {
+	CustomID   ComponentID          `json:"custom_id"`
+	Components *ContainerComponents `json:"components,omitempty"`
```
diamondburned: Is this `null`able? Do we need the pointer type?

diamondburned (changes requested):

api/interaction.go
itslychee: Yeah I was a bit stuck on the naming so I just decided to go with what Discord named it, and I also agree with renaming the constants to `xResponseType`, It would make it easier to know what to look for IMO.

discord/interaction.go
itslychee: Yeah my mistake, Modals require an array of components so let me fix that.

api/interaction.go
itslychee: Sure will

diamondburned (approved):

PR #311, opened 2022-02-14
Add Autocomplete flags to Number and Integer options

tystuyfzand:
Discord's documentation states that Number and Integer types can support autocomplete:

https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure

### DIFF
--- a/discord/command.go
+++ b/discord/command.go
@@ -385,12 +385,13 @@ type StringChoice struct {
 
 // IntegerOption is a subcommand option that fits into a CommandOptionValue.
 type IntegerOption struct {
-	OptionName  string          `json:"name"`
-	Description string          `json:"description"`
-	Required    bool            `json:"required"`
-	Min         option.Int      `json:"min_value,omitempty"`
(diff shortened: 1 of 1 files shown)


discord/command.go
```diff
@@ -385,12 +385,13 @@ type StringChoice struct {
...
-	Min         option.Int      `json:"min_value,omitempty"`
-	Max         option.Int      `json:"max_value,omitempty"`
-	Choices     []IntegerChoice `json:"choices,omitempty"`
+	OptionName   string          `json:"name"`
+	Description  string          `json:"description"`
+	Required     bool            `json:"required"`
+	Min          option.Int      `json:"min_value,omitempty"`
+	Max          option.Int      `json:"max_value,omitempty"`
+	Choices      []IntegerChoice `json:"choices,omitempty"`
+	Autocomplete bool            `json:"autocomplete"`
```
diamondburned: I think we should document that `Autocomplete must not be true if Choices are present.`

diamondburned (approved):

PR #313, opened 2022-02-17
discord: Add Icon/Emoji & Tags to Role

Vendicated:
PremiumSubscriber seems unused for now, the docs even have it typed as "null", but it is already in the docs so i decided to include it

This now also alters api, but I am not sure how to reflect that in the title

### DIFF
--- a/api/role.go
+++ b/api/role.go
@@ -70,6 +70,15 @@ type CreateRoleData struct {
 	// Default: false
 	Mentionable bool `json:"mentionable,omitempty"`
 
+	// Icon is the icon of the role. Requires the guild to have the ROLE_ICONS feature.
+	//
+	// Default: null
+	Icon *Image `json:"icon,omitempty"`
+	// UnicodeEmoji is the role's unicode emoji. Requires the guild to have the ROLE_ICONS feature. This value is nullable.
+	//
+	// Default: null
+	UnicodeEmoji string `json:"unicode_emoji,omitempty"`
+
 	AddRoleData `json:"-"`
 }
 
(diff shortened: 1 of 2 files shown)


api/role.go
```diff
@@ -133,6 +142,11 @@ type ModifyRoleData struct {
 	// Mentionable specifies whether the role should be mentionable.
 	Mentionable option.NullableBool `json:"mentionable,omitempty"`
 
+	// Icon is the icon of the role. Requires the guild to have the ROLE_ICONS feature. This value is nullable.
+	Icon *Image `json:"icon,omitempty"`
+	// UnicodeEmoji is the role's unicode emoji. Requires the guild to have the ROLE_ICONS feature. This value is nullable.
+	UnicodeEmoji string `json:"unicode_emoji,omitempty"`
```
diamondburned: We might need `option.NullableString` here.

api/role.go
```diff
@@ -133,6 +142,11 @@ type ModifyRoleData struct {
 	// Mentionable specifies whether the role should be mentionable.
 	Mentionable option.NullableBool `json:"mentionable,omitempty"`
 
+	// Icon is the icon of the role. Requires the guild to have the ROLE_ICONS feature. This value is nullable.
+	Icon *Image `json:"icon,omitempty"`
```
diamondburned, line 148: How would one unset a role icon? `nil` here would just not update it. Would `&Image{}` work? `**Image` might be needed.

diamondburned (changes requested):

api/role.go
Vendicated: Don't you just have to set the content to an empty array? https://github.com/diamondburned/arikawa/blob/6dc189519c4e96b9ed5e77c9afd37c8617ae771f/api/image.go#L101-L102

so yes &Image{} should do the trick. Perhaps this should be mentioned in the doc comment? (Although I copy pasted this line from the UpdateUserData struct so it would also need to be updated in other places)

diamondburned: Ah, perhaps documenting it would be a better idea. You should document that `NullImage` can be used, and I can add a `var NullImage` in before merging the PR.

Vendicated: Sounds good! Will do that in a bit

diamondburned (approved):

PR #319, opened 2022-03-29
Support guild scheduled events

itslychee:


api/scheduled_events.go
```diff
@@ -0,0 +1,146 @@
...
+	// Member is the member object of the user.
+	Member *discord.Member `json:"member"`
+}
+
+// ListScheduledEventUsers returns a list of users currently in a scheduled event.
+//
+// https://discord.com/developers/docs/resources/guild-scheduled-event#get-guild-scheduled-event-users
+func (c *Client) ListScheduledEventUsers(
+	guildID discord.GuildID, eventID discord.EventID, limit option.NullableInt,
+	withMember bool, before, after discord.UserID) ([]*GuildScheduledEventUser, error) {
```
diamondburned: Return a `[]GuildScheduledEventUser` here for consistency.

api/scheduled_events.go
```diff
@@ -0,0 +1,146 @@
...
+		&eventUsers, "GET", EndpointGuilds+guildID.String()+"/scheduled-events/"+eventID.String()+"/users",
+		httputil.WithSchema(c, params),
+	)
+
+}
+
+// ListScheduledEvents lists the scheduled events in a guild.
+//
+// https://discord.com/developers/docs/resources/guild-scheduled-event#get-guild-scheduled-event-users
+func (c *Client) ListScheduledEvents(guildID discord.GuildID, withUserCount bool) ([]*discord.GuildScheduledEvent, error) {
```
diamondburned: See above.

discord/scheduled_events.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/diamondburned/arikawa/v3/utils/json/option"
+)
+
+// EventStatus describes the different statuses GuildScheduledEvent can be.
+//
+// https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-object-guild-scheduled-event-status
+type EventStatus int
+
+const (
+	Scheduled EventStatus = iota + 1
```
diamondburned: Name these constants `ScheduledEvent`, `ActiveEvent`, etc. instead, so they're easier to read and are less ambiguous.

discord/scheduled_events.go
```diff
@@ -0,0 +1,85 @@
...
+	PrivacyLevel ScheduledEventPrivacyLevel `json:"privacy_level"`
+	// Status is the status of the scheduled event.
+	Status EventStatus `json:"status"`
+	// EntityType describes the type of scheduled event.
+	EntityType EntityType `json:"entity_type"`
+	// EntityID is the id of an entity associated with a scheduled event.
+	EntityID EntityID `json:"entity_id"`
+	// EntityMetadata is additional metadata for the scheduled event.
+	EntityMetadata *EntityMetadata `json:"entity_metadata"`
+	// Creator is the the user responsible for creating the scheduled event.
```
diamondburned: Document that this field is only present (not null) when `CreatorID` is valid.

discord/scheduled_events.go
```diff
@@ -0,0 +1,85 @@
...
+	// to ExternalEntity.
+	ChannelID ChannelID `json:"channel_id"`
+	// CreatorID is the user id of who created the scheduled event.
+	CreatorID UserID `json:"creator_id"`
+	// Name is the name of the scheduled event.
+	Name string `json:"name"`
+	// Description is the description of the scheduled event.
+	Description option.NullableString `json:"description"`
+	// StartTime is when the scheduled event will start at.
+	StartTime time.Time `json:"scheduled_start_time"`
```
diamondburned: Use type `Timestamp`.

discord/scheduled_events.go
```diff
@@ -0,0 +1,85 @@
...
+	// EndTime is when the scheduled event will end at, if it does.
+	EndTime *time.Time `json:"scheduled_end_time"`
```
diamondburned: Also use type `Timestamp`. The user can check if it's valid by using `EndTime.IsValid()`.

discord/scheduled_events.go
```diff
@@ -0,0 +1,85 @@
...
+	// EntityID is the id of an entity associated with a scheduled event.
+	EntityID EntityID `json:"entity_id"`
+	// EntityMetadata is additional metadata for the scheduled event.
+	EntityMetadata *EntityMetadata `json:"entity_metadata"`
+	// Creator is the the user responsible for creating the scheduled event.
+	Creator *User `json:"creator"`
+	// UserCount is the number of users subscribed to the scheduled event.
+	UserCount int `json:"user_count"`
+	// Image is the cover image hash of the scheduled event.
+	Image *Hash `json:"image"`
```
diamondburned: Since the empty string value of `Image` is meaningless and there's no known difference between a `null` Image and an omitted one, this field should be `Image Hash` with an `omitempty` tag.

discord/scheduled_events.go
```diff
@@ -0,0 +1,85 @@
...
+	// ChannelID is the channel id in which the scheduled event will be
+	// hosted at, this may be NullChannelID if the EntityType is set
+	// to ExternalEntity.
+	ChannelID ChannelID `json:"channel_id"`
+	// CreatorID is the user id of who created the scheduled event.
+	CreatorID UserID `json:"creator_id"`
+	// Name is the name of the scheduled event.
+	Name string `json:"name"`
+	// Description is the description of the scheduled event.
+	Description option.NullableString `json:"description"`
```
diamondburned: No difference between `null` and omitted here, so use a `string` with an `omitempty` flag.

diamondburned (changes requested):

diamondburned (approved):

PR #321, opened 2022-04-11
discord: Add localization to Application Commands

Rark-ness:
* Available languages can be found in this document:  https://discord.com/developers/docs/reference#locales
* Also, refer to this document for application command localization: https://discord.com/developers/docs/interactions/application-commands#localization

I added all the fields to the struct.
However, I am not sure how it would be efficient to add localization to functions such as `NewStringOption()`. Please give me some advice.

### DIFF
--- a/utils/ws/op.go
+++ b/utils/ws/op.go
@@ -18,7 +18,7 @@ type OpCode int
 type CloseEvent struct {
 	// Err is the underlying error.
 	Err error
-	// Code is the websocket close code, if any.
+	// Code is the websocket close code, if any. It is -1 otherwise.
 	Code int
 }
 


discord/command.go
```diff
@@ -53,6 +59,39 @@ type Command struct {
 	Version Snowflake `json:"version,omitempty"`
 }
 
+type CommandLocales struct {
```
diamondburned: I think it would be best if you implement this as

```go
type CommandLocales map[Language]string

// Language is a type that describes a language code such as "de" or "en-US".
type Language string

const (
	EnglishUS Language = "en-US"
	EnglishUK Language = "en-UK"
	Spanish   Language = "es-ES"
	// ...
)
```

This will also make `,omitempty` actually work.

discord/command.go
```diff
@@ -33,9 +33,15 @@ type Command struct {
...
-	Name string `json:"name"`
+	Name              string         `json:"name"`
+	NameLocalizations CommandLocales `json:"name_localizations,omitempty"`
 	// Description is the 1-100 character description.
-	Description string `json:"description"`
+	Description              string         `json:"description"`
+	DescriptionLocalizations CommandLocales `json:"description_localizations,omitempty"`
+	// The fields NameLocalized, DescriptionLocalized are
+	// populated only when data is received.
+	NameLocalized        string `json:"name_localized,omitempty"`
```
diamondburned: These are better named `LocalizedName` and `LocalizedDescription`, I think. The JSON tags stay the same.

diamondburned (changes requested):
I don't think there's a way to specify the `locale` field yet, so I think the `LocalizedName` and `LocalizedDescription` fields will never be populated. Can you double-check that?

Rark-ness:
The official document explains as below, so I think the value will come in through the user's language without setting the header:
> Locale is determined by looking at the `X-Discord-Locale` header, then the `Accept-Language` header if not present, then lastly the user settings locale.

However, I think we need a separate function to set the header.

diamondburned:
> However, I think we need a separate function to set the header.

Just to keep the API simple, I think we can add a

```go
func  (*api.Client) WithLocale(language discord.Language) *api.Client
```

discord/command.go
```diff
@@ -53,6 +59,43 @@ type Command struct {
 	Version Snowflake `json:"version,omitempty"`
 }
 
+type Language string
```
diamondburned: Can you add the needed comments?

PR #326, opened 2022-05-06
api/discord: add DefaultMemberPermissions and NoDMPermission to commands

Juby210:


discord/command.go
```diff
@@ -52,6 +52,11 @@ type Command struct {
 	//
 	// It is only present on ChatInputCommands.
 	Options CommandOptions `json:"options,omitempty"`
+	// DefaultMemberPermissions is set of permissions.
+	DefaultMemberPermissions Permissions `json:"default_member_permissions,string,omitempty"`
+	// DmPermission indicates whether the command is available in DMs with
+	// the app, only for globally-scoped commands. By default, commands are visible.
+	DmPermission bool `json:"dm_permission,omitempty"`
```
topi314: This fields defaults to true
Perhaps it should be inverted like `NoDefaultPermission`

discord/command.go
```diff
@@ -52,6 +52,11 @@ type Command struct {
 	//
 	// It is only present on ChatInputCommands.
 	Options CommandOptions `json:"options,omitempty"`
+	// DefaultMemberPermissions is set of permissions.
+	DefaultMemberPermissions Permissions `json:"default_member_permissions,string,omitempty"`
```
topi314: 0 is a valid permission here you can set to disallow anyone except it they have admin or an override to use the command, if it's 0 it gets omitted perhaps thats not ideal

Juby210: I see, do you have any proposition to fix this?

topi314: I guess a pointer would do the job, not sure if it is the best solution here
how does arikawa does it for other fields like this?

api/application.go
```diff
@@ -26,6 +26,8 @@ type CreateCommandData struct {
 	Description              string                 `json:"description"`
 	DescriptionLocalizations discord.StringLocales  `json:"description_localizations,omitempty"`
 	Options                  discord.CommandOptions `json:"options,omitempty"`
+	DefaultMemberPermissions discord.Permissions    `json:"default_member_permissions,string,omitempty"`
+	NoDmPermission           bool                   `json:"-"`
```
diamondburned: This should be `NoDMPermission`.

api/application.go
```diff
@@ -34,13 +36,15 @@ func (c CreateCommandData) MarshalJSON() ([]byte, error) {
 	type RawCreateCommandData CreateCommandData
 	cmd := struct {
 		RawCreateCommandData
+		DmPermission      bool `json:"dm_permission"`
```
diamondburned: `DMPermission`

discord/command.go
diamondburned: You can use `option.Uint`, but `*Permissions` can work, too. I would make a helper function `NewPermissions(p ...Permissions) *Permissions` in that case.

diamondburned:
Above changes must be done before this PR can be merged.

diamondburned:
Also, please link the documentation for these 2 fields in the code.

Juby210:
> Also, please link the documentation for these 2 fields in the code.

Documentation is linked to struct itself, so I guess this is enough.
Also applied your suggestions.

diamondburned (approved):

PR #329, opened 2022-05-31
Autocomplete Options changes

twoscott:
Currently AutocompleteOption behaves as if the option can only be a string. This makes sense since only StringOption can be autocompleted, but AutocompleteInteraction provides all options, not just the focused StringOption that is being typed in. So the way it's currently written is limiting and would mean you need to convert from string to a given type if you want to use any of the other options' values in your autocomplete logic. This PR mimics the behaviour of CommandInteractionOption, not sure if there's a more elegant way than duplicating the methods to AutocompleteOption. These changes are breaking so not sure if you want it in v3.

### DIFF
--- a/0-examples/autocomplete/main.go
+++ b/0-examples/autocomplete/main.go
@@ -42,7 +42,7 @@ func main() {
 				},
 			}
 		case *discord.AutocompleteInteraction:
-			allChoices := []api.AutocompleteChoice{
+			allChoices := api.AutocompleteStringChoices{
 				{Name: "Choice A", Value: "Choice A"},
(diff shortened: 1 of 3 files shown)


twoscott:
Commit https://github.com/diamondburned/arikawa/pull/329/commits/6b71377dc5ddcb30a58a458e5f295d28935c5b3a allows for all three of the documented choice value types to be used when sending autocomplete result interaction responses. These types are [String, Integer, and Number](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-choice-structure).

diamondburned (approved):

PR #331, opened 2022-07-09
Add AttachmentOption type

twoscott:
This PR adds a missing command option type which lets users attach files as a slash command option.
https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-type

### DIFF
--- a/discord/command.go
+++ b/discord/command.go
@@ -310,6 +310,7 @@ const (
 	RoleOptionType
 	MentionableOptionType
 	NumberOptionType
+	AttachmentOptionType
 	maxOptionType // for bound checking
 )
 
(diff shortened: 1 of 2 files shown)


discord/command.go
```diff
@@ -328,6 +329,7 @@ const (
 //    - *RoleOption
 //    - *MentionableOption
 //    - *NumberOption
+//	  - *AttachmentOption
```
diamondburned, line 332: Seems to be a missing space here.

twoscott: Ah I think it's just that I used a tab and then 2 spaces, I'll swap it for spaces.

diamondburned (approved):

Issue #333, opened 2022-07-15
Improperly named JSON fields in TextInputComponent


DemmyDemon:
[These values](https://github.com/diamondburned/arikawa/blob/v3/discord/component.go#L501-L502) are named wrong.

According to [the Discord documentation](https://discord.com/developers/docs/interactions/message-components#text-inputs-text-input-example), they should be named `min_length` and `max_length`.

diamondburned:
Huh. How the hell did this even happen?

PR #334, opened 2022-07-16
Fix TextInputComponent JSON fields

twoscott:
`min_values` and `max_values` should be `min_length` and `max_length` for TextInputComponent as defined in #333 and the [documentation](https://discord.com/developers/docs/interactions/message-components#text-inputs-text-input-structure).

### DIFF
--- a/discord/component.go
+++ b/discord/component.go
@@ -471,8 +471,8 @@ type TextInputComponent struct {
 	Style TextInputStyle `json:"style"`
 	// Label is the title of this component, describing its use
 	Label string `json:"label"`
-	// ValueLimits is the minimum and maximum length for the input
-	ValueLimits [2]int `json:"-"`
+	// LengthLimits is the minimum and maximum length for the input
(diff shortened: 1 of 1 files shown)


diamondburned (approved):

discord/component.go
```diff
@@ -508,11 +508,11 @@ func (i *TextInputComponent) MarshalJSON() ([]byte, error) {
 	}
 
 	if i.ValueLimits != [2]int{0, 0} {
```
diamondburned: We should rename this to `LengthLimits` then.

diamondburned (changes requested):

discord/component.go
twoscott: That's true I missed that. Resolved with https://github.com/diamondburned/arikawa/pull/334/commits/c36309ed0db0dba155ae0830e77a52c15c216a78

diamondburned (approved):

PR #335, opened 2022-07-20
Name Member guild ID params appropriately

twoscott:
The parameter that needs to be passed to these Member.AvatarURL*() methods is a guild ID but the parameters are only named `guild`. This change renames the parameters to `guildID` - making the required parameters clearer at a glance.


diamondburned (approved):
I personally think the type name already makes that painfully obvious, but this change causes no harm.

PR #336, opened 2022-07-22
Add Invite.URL() methods

twoscott:
This PR adds two convenience methods for `Invite` that allow users to easily access the URL of an invite, similarly to `Message.URL()` and `Emoji.URL()` etc.

### DIFF
--- a/discord/invite.go
+++ b/discord/invite.go
@@ -32,6 +32,16 @@ type Invite struct {
 	InviteMetadata
 }
 
+// URL returns a Discord invite URL linking to the invite.
+func (i Invite) URL() string {
+	return "https://discord.gg/" + i.Code
+}
+
+// LongURL returns a long-form Discord invite URL linking to the invite.
+func (i Invite) LongURL() string {
+	return "https://discord.com/invite/" + i.Code
+}
+
 // https://discord.com/developers/docs/resources/invite#invite-object-target-user-types
 type InviteUserType uint8
 


discord/invite.go
```diff
@@ -32,6 +32,16 @@ type Invite struct {
 	InviteMetadata
 }
 
+// URL returns a Discord invite URL linking to the invite.
+func (i Invite) URL() string {
+	return "https://discord.gg/" + i.Code
```
diamondburned, line 37: Can we have a constant for this URL?

discord/invite.go
```diff
@@ -32,6 +32,16 @@ type Invite struct {
...
+}
+
+// LongURL returns a long-form Discord invite URL linking to the invite.
+func (i Invite) LongURL() string {
+	return "https://discord.com/invite/" + i.Code
```
diamondburned, line 42: Use `BaseEndpoint` for this.

diamondburned (changes requested):

discord/invite.go
twoscott: Yeah I didn't think about that. Which package do you think it should be defined in? 

Should we also use constants for `cdn.discordapp` URLs too? e.g. [`emoji.go`](https://github.com/diamondburned/arikawa/blob/v3/discord/emoji.go#L87) 
There's actually quite a few instances of plain-text `cdn.discordapp` URLs scattered around.

discord/invite.go
twoscott: Should that const also be used here then? [`message.go`](https://github.com/diamondburned/arikawa/blob/v3/discord/message.go#L118)

diamondburned: If you want an `EndpointInvite`, then it should probably be in this file. `BaseEndpoint + "/invite/"` is OK, though.

diamondburned: Actually, `EndpointInvite` should probably be in invite.go.

discord/invite.go
diamondburned: invite.go is good.

> Should we also use constants for `cdn.discordapp` URLs too? e.g. [`emoji.go`](https://github.com/diamondburned/arikawa/blob/v3/discord/emoji.go#L87)
> There's actually quite a few instances of plain-text `cdn.discordapp` URLs scattered around.

That'd be great, but it can be a separate PR.

discord/invite.go
diamondburned: Resolving these 2 conversations, since I don't really want additional URL constants in `package discord`. In hindsight, having all URL constants in `discord` would've been the better idea rather than `api`.

diamondburned (approved):

PR #349, opened 2022-10-03
fix(gateway): make RequestGuildMembersCommand.Query and Limit nullable

starshine-sys:


gateway/events.go
```diff
@@ -76,10 +77,10 @@ type RequestGuildMembersCommand struct {
 	GuildIDs []discord.GuildID `json:"guild_id"`
 	UserIDs  []discord.UserID  `json:"user_ids,omitempty"`
 
-	Query     string `json:"query,omitempty"`
-	Limit     uint   `json:"limit,omitempty"`
-	Presences bool   `json:"presences"`
-	Nonce     string `json:"nonce,omitempty"`
+	Query     option.NullableString `json:"query"`
```
diamondburned: Query should have `,omitempty`; it's an optional field: https://discord.com/developers/docs/topics/gateway-events#request-guild-members.

gateway/events.go
```diff
@@ -76,10 +77,10 @@ type RequestGuildMembersCommand struct {
...
+	Limit     option.NullableUint   `json:"limit"`
```
diamondburned: Limit should also have `,omitempty` just because the official client sometimes excludes it. Can we check if omitting this field will still make them work for bots?

diamondburned: If not, an idea would be to not omit it unless `Limit == -1`, then we can explicitly document them for non-bot users. This way, we prioritize bot support.

diamondburned (changes requested):

starshine-sys:
The command was fixed in d97492348d7c3ce174d2b66ff29ae015ab84809d, so this PR isn't necessary anymore

PR #350, opened 2022-10-11
feat: add support for disabling signature verification

zBNF:
During local development it doesn't make a lot of sense to require signature verification.

Here's a proposed change that allows library users to disable verification. I'm open to alternative approaches in order to avoid introducing any breaking changes if there are concerns around that.

### DIFF
--- a/api/webhook/interactionserver.go
+++ b/api/webhook/interactionserver.go
@@ -87,12 +87,11 @@ type InteractionServer struct {
 	interactionHandler InteractionHandler
 	httpHandler        http.Handler
 	pubkey             ed25519.PublicKey
-	doVerification     bool
 }
 
 // NewInteractionServer creates a new InteractionServer instance. pubkey should
 // be hex-encoded.
-func NewInteractionServer(pubkey string, handler InteractionHandler, doVerification bool) (*InteractionServer, error) {
+func NewInteractionServer(pubkey string, handler InteractionHandler) (*InteractionServer, error) {
 	pubkeyB, err := hex.DecodeString(pubkey)
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot decode hex pubkey")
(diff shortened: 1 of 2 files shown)


diamondburned:
Discord won't allow using a new endpoint without having the valid signature in the first place.

I'm not sure what you're actually trying to accomplish by having this PR. I personally develop my bot just fine with this. Are you trying to mock test the bot? Why not just call the handler function directly?

diamondburned:
A non-breaking way to introduce this would be to not do verification if `pubkey == ""`. This way, you can empty out the environment variable for the pubkey within development. No new parameters need to be introduced.

zBNF:
I think this works.

> Are you trying to mock test the bot?

Basically, I'm looking to run black box tests using mocks on my binary.

api/webhook/interactionserver.go
```diff
@@ -107,14 +107,20 @@ func NewInteractionServer(pubkey string, handler InteractionHandler) (*Interacti
...
+	if len(s.pubkey) != 0 {
+		s.httpHandler = s.withVerification(s.httpHandler)
+	}
 
 	return &s, nil
 }
 
 // ServeHTTP implements http.Handler.
 func (s *InteractionServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	s.withVerification(http.HandlerFunc(s.handle)).ServeHTTP(w, r)
```
diamondburned, line 117: I think `ServeHTTP` should just be calling `s.httpHandler(w, r)`. I must've introduced the variable but forgot to use it.

api/webhook/interactionserver.go
```diff
@@ -107,14 +107,20 @@ func NewInteractionServer(pubkey string, handler InteractionHandler) (*Interacti
 	}
 
 	s.httpHandler = http.HandlerFunc(s.handle)
-	s.httpHandler = s.withVerification(s.httpHandler)
+	if len(s.pubkey) != 0 {
```
diamondburned, line 110: Change to `s.pubkey != ""`.

diamondburned (changes requested):

api/webhook/interactionserver.go
zBNF: That's invalid.
```invalid operation: cannot compare s.pubkey != "" (mismatched types ed25519.PublicKey and untyped string)```

diamondburned (approved):

PR #353, opened 2022-11-19
discord: Add new select components

twoscott:
This pull request adds the [new select components](https://discord.com/developers/docs/change-log#new-select-menu-components). It also renames `SelectComponent` to `StringSelectComponent` in-line with the [documentation](https://discord.com/developers/docs/interactions/message-components#component-object-component-types). `TextInputComponent.Value` was also changed from `option.NullableString` to `string` as I don't see a reason for it to be optional - when you submit from a modal, the values in the text input components are sent to the API and then received as a `MESSAGE_COMPONENT` interaction event where the attached components have the same values submitted into the modal, so I can't see where it would need to be nullable?

### DIFF
--- a/discord/component.go
+++ b/discord/component.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/diamondburned/arikawa/v3/internal/rfutil"
 	"github.com/diamondburned/arikawa/v3/utils/json"
-	"github.com/diamondburned/arikawa/v3/utils/json/option"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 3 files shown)


diamondburned:
> so I can't see where it would need to be nullable

The documentation doesn't seem to suggest this, so I agree.

Are there any other components that still use `option.NullableString`? Are the examples updated?

twoscott:
No other components use `option.NullableString`, and the examples are updated [here](https://github.com/diamondburned/arikawa/pull/353/commits/cb129ee73334ffa202a3658d2048f43a22504129#diff-0bff85917e667a8a0a0ab0f490d881484f58eae6335ceb50904d3346d9063bcb).

diamondburned (approved):

PR #354, opened 2022-11-19
discord: Add StringOption Min/MaxLength

twoscott:
This pull request adds Min/MaxLength fields for `StringOption`, documented [here](https://discord.com/developers/docs/change-log#min-and-max-length-for-command-options). Min/Max fields for `IntegerOption` and `NumberOption` were also renamed to Min/MaxValue to differentiate better between `...Value` and `...Length` fields. This is breaking but makes them more accurate to the API and there may be breaking changes in #353 anyway.

### DIFF
--- a/discord/command.go
+++ b/discord/command.go
@@ -447,6 +447,8 @@ type StringOption struct {
 	DescriptionLocalizations StringLocales  `json:"description_localizations,omitempty"`
 	Required                 bool           `json:"required"`
 	Choices                  []StringChoice `json:"choices,omitempty"`
+	MinLength                option.Int     `json:"min_length,omitempty"`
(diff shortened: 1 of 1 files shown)


diamondburned:
> Min/Max fields for `IntegerOption` and `NumberOption` were also renamed to Min/MaxValue to differentiate better between `...Value` and `...Length` fields. This is breaking but makes them more accurate to the API and there may be breaking changes in #353 anyway.

I don't think this is necessary. Min and Max makes perfect sense given an integer/number type.

twoscott:
That's true, amended in [f62a501](https://github.com/diamondburned/arikawa/pull/354/commits/f62a50123c97c9bbaf491c84ac8c2c6e8ab9224c).

diamondburned (approved):

PR #357, opened 2022-11-28
Add omitempty tag for category id

XSAM:
Without this, the category id field is a required parameter when requesting `MoveChannel` even though all I want to do is move the position of the channel.

This required parameter also causes an error when I try to move multiple channels at once.

```
Discord 400 error: Only one channel can have a parent_id modified at a time
```


diamondburned (approved):

diamondburned:
I accidentally tagged this commit wrong. Whoops.

PR #358, opened 2022-12-04
Close previous UDP voice sessions on Continue

1lann:
Previously, stale UDP voice sessions (i.e. when the region changes or the voice server fails over and a server change notification is sent via the websocket) were never closed, this could leave callers hanging perpetually unless they set a Read/Write timeout. However that's kind of an annoying quirk that can be worked around by just closing the UDP session when appropriate. I'm not sure if `Continue` is the right place to close the connection, but it works for me lol. There is probably a more correct location to close the connection? I'm just making this PR to give you an idea I guess.

### DIFF
--- a/voice/udp/manager.go
+++ b/voice/udp/manager.go
@@ -28,6 +28,7 @@ type Manager struct {
 
 	// conn state
 	conn     *Connection
+	prevConn *Connection
 	connLock chan struct{}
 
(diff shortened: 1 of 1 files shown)


diamondburned (approved):

diamondburned:
I have no clue what this is really doing, but if it works, it works.

LGTM.

PR #369, opened 2023-01-26
Add Flags field to Member struct

ayn2op:
Reference: https://github.com/discord/discord-api-docs/pull/5874


diamondburned (approved):

diamondburned:
It turns out `MemberFlags` can be larger than a `uint8`!

```
23:08:00.695 INF warning: gateway: background gateway error: cannot unmarshal JSON event (op 0 type "READY") from gateway: json: cannot unmarshal number 522 into Go struct field Member.guilds.members.flags of type discord.MemberFlags
```

PR #370, opened 2023-01-27
Rename GuildBans intent & add GUILD_AUDIT_LOG_ENTRY_CREATE event

ayn2op:
Reference: https://discord.com/developers/docs/change-log#guild-audit-log-events

### DIFF
--- a/gateway/intents.go
+++ b/gateway/intents.go
@@ -12,7 +12,7 @@ type Intents uint32
 const (
 	IntentGuilds Intents = 1 << iota
 	IntentGuildMembers
-	IntentGuildBans
+	IntentGuildModeration
 	IntentGuildEmojis
 	IntentGuildIntegrations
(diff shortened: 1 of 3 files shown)


gateway/intents.go
```diff
@@ -12,7 +12,7 @@ type Intents uint32
 const (
 	IntentGuilds Intents = 1 << iota
 	IntentGuildMembers
-	IntentGuildBans
```
diamondburned, line 15: Can we add an alias to avoid breaking old code?

```go
// IntentGuildBans aliases to IntentGuildModeration.
//
// Deprecated: IntentGuildModeration is the more correct constant to use.
const IntentGuildBans = IntentGuildModeration

diamondburned (changes requested):

diamondburned (approved):

PR #372, opened 2023-02-01
fix IntegerOptionType Unmarshall

bitspill:
All numerical types were using `NumberOptionType` rather than distinguishing `IntegerOptionType` which led to an unmarshalling error when using the int/uint variants

`/x 12` -> `option "x" expecting type 10, got 4`
```go
// Initialize bot connection
// Register a command for x
// Register route for x to cmdX

api.CreateCommandData{
	Name:        "x",
	Description: "x",
	&discord.IntegerOption{
		OptionName:  "x",
		Description: "Integer x",
		Required:    true,
	}
}

func (b *Bot) cmdX(ctx context.Context, data cmdroute.CommandData) *api.InteractionResponseData {
	var options struct {
		X int `discord:"x"`
	}

	if err := data.Options.Unmarshal(&options); err != nil {
		return errorResponse(err)
	}
	return &api.InteractionResponseData{
		Content:         option.NewNullableString(fmt.Sprintf("X was (%d)", options.X)),
		AllowedMentions: &api.AllowedMentions{}, // don't mention anyone
	}
}
```


diamondburned (approved):

diamondburned (changes requested):
Actually, before I merge this PR, I think this is a breaking change. Can we modify the code to use a `switch` that accepts both types instead of using a `map`?

bitspill:
Ah yes good catch and idea, I'll do that when my current project wraps up in the next couple weeks.

diamondburned:
Replaced by #421.

PR #375, opened 2023-02-12
refactor(cmdroute)!: change data to resp

Serpentiel:
changes `cmdroute` to work with `*api.InteractionResponse` instead of `*api.InteractionResponseData`

this would allow to return modals via command interactions done through `cmdroute`

this PR is a breaking change

### DIFF
--- a/0-examples/commands-hybrid/main.go
+++ b/0-examples/commands-hybrid/main.go
@@ -109,13 +109,16 @@ func newHandler(s *state.State) *handler {
 	return h
 }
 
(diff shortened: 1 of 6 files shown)


zwinnerman-fleetdm (approved):

diamondburned:
I think it would be better if we introduce a new non-breaking API that apps can use to implement this when they need to instead of breaking the current API.

I'll have to look into this soon.

Serpentiel:
to make it non-breaking, I believe one has to modify this place: https://github.com/diamondburned/arikawa/blob/181dcb1bdd007d4d8f42c744278f8f59265bb2b4/api/cmdroute/router.go#L123-L127

and introduce a new type of handler which is used there then

PR #393, opened 2023-06-06
Add support for the new username system

starshine-sys:
Adds what is currently documented in https://github.com/discord/discord-api-docs/pull/6130. The documentation isn't merged yet, but this has already started rolling out to normal users.

### DIFF
--- a/discord/user.go
+++ b/discord/user.go
@@ -8,8 +8,9 @@ import (
 type User struct {
 	ID            UserID `json:"id"`
 	Username      string `json:"username"`
-	Discriminator string `json:"discriminator"`
+	Discriminator string `json:"discriminator"` // This is "0" if the user has migrated to the new username system.
 	Avatar        Hash   `json:"avatar"`
+	GlobalName    string `json:"global_name"`
 
 	// These fields may be omitted
 
@@ -42,6 +43,10 @@ func (u User) Mention() string {
 
 // Tag returns a tag of the user.
 func (u User) Tag() string {
+	if u.Discriminator == "0" {
+		return u.Username
+	}
+
 	return u.Username + "#" + u.Discriminator
 }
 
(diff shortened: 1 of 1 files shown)


diamondburned:
Is the `GlobalName` here the display name? AFAIK the new global username is kept the same in the `Username` field?

starshine-sys:
Yeah, `GlobalName` is the display name--I'm pretty sure the field was originally called `display_name` but it got changed before the rollout.

diamondburned:
> I'm pretty sure the field was originally called `display_name` but it got changed before the rollout.

Is there a good reason for this? Should we call it `DisplayName` on our end (to match the client) or keep? Does Discord intend to eventually remove display names?

starshine-sys:
I can't think of any good reason for it being changed other than maybe reducing confusion with server nicknames, so `DisplayName` is probably a better name for it, yeah. I doubt Discord will remove display names because it's prominently shown in the UI and mentioned in all their support articles.

diamondburned:
Can you make the change? I'll approve the PR afterwards.

discord/user.go
```diff
@@ -8,8 +8,9 @@ import (
 type User struct {
 	ID            UserID `json:"id"`
 	Username      string `json:"username"`
-	Discriminator string `json:"discriminator"`
+	Discriminator string `json:"discriminator"` // This is "0" if the user has migrated to the new username system.
 	Avatar        Hash   `json:"avatar"`
+	GlobalName    string `json:"global_name"`
```
diamondburned: `s/GlobalName/DisplayName/`

diamondburned (changes requested):

diamondburned (approved):

diamondburned:
In hindsight, `GlobalName` would've been better, since that would allow a `DisplayName() string` method. I will add a `DisplayOrUsername() string` though.

PR #411, opened 2023-11-03
state: Use store roles for calculating overrides

twoscott:
Currently, the state.Permissions() helper method uses Guild.Roles to calculate the permission overrides. Guild.Roles aren't updated by Guild Role Update events, so this can result in the helper method giving unexpected results when a role was updated, but no Guild Update event has happened since. This change pulls roles from store.Roles() in the helper method, and passes them to the CalcOverrides method, which now uses the passed in roles instead of Guild.Roles.

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -299,15 +299,18 @@ func (s *State) Permissions(
 	var wg sync.WaitGroup
 
 	var (
-		g *discord.Guild
-		m *discord.Member
+		g  *discord.Guild
+		m  *discord.Member
+		rs []discord.Role
 
 		gerr = store.ErrNotFound
 		merr = store.ErrNotFound
+		rerr = store.ErrNotFound
 	)
 
 	if s.HasIntents(gateway.IntentGuilds) {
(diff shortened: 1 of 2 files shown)


state/state.go
```diff
@@ -793,6 +809,17 @@ func (s *State) fetchGuild(id discord.GuildID) (g *discord.Guild, err error) {
 	return
 }
 
+func (s *State) fetchRoles(gID discord.GuildID) (rs []discord.Role, err error) {
+	rs, err = s.Session.Roles(gID)
+	if err == nil && s.HasIntents(gateway.IntentGuilds) {
```
diamondburned, line 814: With this check, we're better off just returning in `Roles()` if the intents are not there. Then, this check simply becomes `if err == nil {`.

The same should also apply for `fetch{Guild,Member}`.

diamondburned (changes requested):

state/state.go
diamondburned: Nevermind, I misread the code. Carry on.

diamondburned (approved):

Issue #422, opened 2024-03-19
Groups do not use autocompleter


patyhank:
CommandHandler handled groups, but AutoCompleter not.

When I use Groups, although I have successfully added an AddAutocompleterFunc without any error
but when I receive an AutocompleteInteraction, it is not handled correctly.

PR #423, opened 2024-03-19
fix cmdroute Autocompleter handling

patyhank:
Fixing #422

### DIFF
--- a/api/cmdroute/router.go
+++ b/api/cmdroute/router.go
@@ -341,13 +341,17 @@ func (r *Router) findAutocompleterOnce(ev *discord.InteractionEvent, data discor
 	if !ok {
 		return autocompleterData{}, false
 	}
-
 	switch node := node.(type) {
 	case routeNodeSub:
 		if len(data.Options) != 1 || data.Type != discord.SubcommandGroupOptionType {
 			break
 		}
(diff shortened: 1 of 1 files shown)


api/cmdroute/router.go
```diff
@@ -316,18 +316,42 @@ type autocompleterData struct {
...
+// findAutocompleter finds the autocomplete handler for the given option name.
+// It checks the current router and its groups.
 func (r *Router) findAutocompleter(ev *discord.InteractionEvent, data discord.AutocompleteOption) (autocompleterData, bool) {
+	found, ok := r.findAutocompleterOnce(ev, data)
+	if ok {
+		return found, true
+	}
+
+	for _, sub := range r.groups {
+		found, ok = sub.findAutocompleterOnce(ev, data)
```
diamondburned, line 328: Should this not be `sub.findAutocompleter()` instead?

diamondburned (changes requested):

api/cmdroute/router.go
patyhank: Does commandHandler have the same problem?

diamondburned: Ah, I see what you mean! This works then, the recursive call is just done elsewhere.

diamondburned (approved):

PR #450, opened 2024-09-19
feat(api): parse Retry-After header as float

### LINKED ISSUE #153: Gateway/API: v2: Millisecond accurate Retry-After by reading the JSON body
v8 changes now have seconds in the `Retry-After` header instead of milliseconds, meaning the rate limiting is now less accurate. We need to read the JSON body for the seconds as a floating point number.

The Limiter API needs to read a body as well now.

ayn2op:
Closes #153

### DIFF
--- a/api/rate/rate.go
+++ b/api/rate/rate.go
@@ -208,12 +208,12 @@ func (l *Limiter) Release(path string, headers http.Header) error {
 
 	switch {
 	case retryAfter != "":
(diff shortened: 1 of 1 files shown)


diamondburned:
What's wrong with this PR?

ayn2op:
> What's wrong with this PR?

According to the documentation, the `float` value is in the response body (the `retry_after` field), not the `Retry-After` response header.

PR #453, opened 2024-10-05
Add new type field

hi117:
https://discord.com/developers/docs/resources/message#message-reference-structure

### DIFF
--- a/discord/message.go
+++ b/discord/message.go
@@ -365,6 +365,13 @@ func (m MessageApplication) CreatedAt() time.Time {
 	return m.ID.Time()
 }
 
+const (
+	// A standard reference used by replies.
+	MessageReferenceTypeDefault = iota
+	// Reference used to point to a message at a point in time.
+	MessageReferenceTypeForward
+)
+
 // MessageReference is used in four situations:
 //
 // # Crosspost messages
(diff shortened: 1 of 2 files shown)


discord/message.go
```diff
@@ -23,6 +23,9 @@ type Message struct {
 	// Flags are the MessageFlags.
 	Flags MessageFlags `json:"flags"`
 
+	// The message associated with the message_reference. This is a minimal subset of fields in a message (e.g. author is excluded.)
+	MessageSnapshots []Message `json:"message_snapshots,omitempty"`
```
diamondburned: Is Author always omitted? If it is, please make a copy of the Message type named MessageSnapshot or something with exactly the fields present.

discord/message.go
```diff
@@ -365,6 +368,16 @@ func (m MessageApplication) CreatedAt() time.Time {
 	return m.ID.Time()
 }
 
+const (
+	// A standard reference used by replies.
+	MessageReferenceTypeDefault = iota
```
diamondburned: Should this be named `MessageReferenceTypeReply` instead?

diamondburned (changes requested):

discord/message.go
ayn2op: @diamondburned Discord API defines type `1` as the default message reference type: https://discord.com/developers/docs/resources/message#message-reference-types

diamondburned: We've had other constants in Arikawa named differently from the documentation just because it's more obvious that way. I think this is one of those cases.

diamondburned: This constant is also untyped. This should be fixed.

discord/message.go
ayn2op: Reference: https://discord.com/developers/docs/resources/message#message-snapshot-object

diamondburned: Wait, this field is wrong then. It's

```go
MessageSnapshots []struct{ Message *Message } // pseudo-code
```

and not

```go
MessageSnapshots []Message
```

discord/message.go
diamondburned: This isn't resolved yet, but I'm choosing to keep the name as `Default`. I'll follow this up with a separate commit.

diamondburned (approved):

PR #455, opened 2024-10-24
discord: Added SelectedValue fields for Channel, Mentionable, User and Role SelectComponents

CosmicPredator:
This PR adds the following fields to ChannelSelectComponent, MentionableSelectComponent and RoleSelectComponent,

- `SelectedRoles` for `RoleSelectComponent`.
- `SelectedChannels` for `ChannelSelectComponent`.
- `SelectedMentions` for `MentionableSelectComponent`.
- `SelectedUsers` for `UserSelectComponent`.

All the fields will take in the slice of ID's of their kind and populates the SelectComponent beforehand.

> [!NOTE]
> `SelectedMentions` will take in the slice of `interface{}` and later type-asserted to respective UserID / RoleID.
> If wanted to change this implementation, please lmk.

### DIFF
--- a/discord/component.go
+++ b/discord/component.go
@@ -720,6 +720,8 @@ type UserSelectComponent struct {
 	ValueLimits [2]int `json:"-"`
 	// Disabled disables the select if true.
 	Disabled bool `json:"disabled,omitempty"`
+	// SelectedUsers is the slice of UserIDs that are marked as selected by default
+	SelectedUsers []UserID `json:"-"`
 }
 
 // ID implements the Component interface.
@@ -737,17 +739,33 @@ func (s *UserSelectComponent) _icp() {}
 func (s *UserSelectComponent) MarshalJSON() ([]byte, error) {
 	type sel UserSelectComponent
 
+	type DefaultValue struct {
+		Id UserID `json:"id"`
+		Type string `json:"type"`
+	}
+
 	type Msg struct {
 		Type ComponentType `json:"type"`
 		*sel
 		MinValues *int `json:"min_values,omitempty"`
 		MaxValues *int `json:"max_values,omitempty"`
+		DefaultValues []DefaultValue `json:"default_values,omitempty"`
 	}
 
 	msg := Msg{
 		Type: UserSelectComponentType,
 		sel:  (*sel)(s),
 	}
+
+	var defaultValues []DefaultValue
+
+	if len(s.SelectedUsers) > 0 {
+		for _, userId := range s.SelectedUsers {
+			defaultValues = append(defaultValues, DefaultValue{Id: userId, Type: "user"})
+		}
+	}
+
+	msg.DefaultValues = defaultValues
 
 	if s.ValueLimits != [2]int{0, 0} {
 		msg.MinValues = new(int)
@@ -840,8 +858,9 @@ type MentionableSelectComponent struct {
 	ValueLimits [2]int `json:"-"`
 	// Disabled disables the select if true.
 	Disabled bool `json:"disabled,omitempty"`
-	// SelectedUsers is the slice of UserIDs that are marked as selected by default
-	SelectedUsers []UserID `json:"-"`
+	// DefaultMentions is the slice of discord.UserID's and discord.RoleID's 
+	// that are marked as selected by default
+	SelectedMentions []interface{} `json:"-"`
 }
 
 // ID implements the Component interface.
@@ -860,7 +879,7 @@ func (s *MentionableSelectComponent) MarshalJSON() ([]byte, error) {
 	type sel MentionableSelectComponent
 
 	type DefaultValue struct {
-		Id UserID `json:"id"`
+		Id Snowflake `json:"id"`
 		Type string `json:"type"`
 	}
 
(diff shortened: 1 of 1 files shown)


discord/component.go
```diff
@@ -720,6 +720,8 @@ type UserSelectComponent struct {
 	ValueLimits [2]int `json:"-"`
 	// Disabled disables the select if true.
 	Disabled bool `json:"disabled,omitempty"`
+	// SelectedUsers is the slice of UserIDs that are marked as selected by default
+	SelectedUsers []UserID `json:"-"`
```
diamondburned: I think this should be named `DefaultUsers` to keep it close to the official documentation. [1]

discord/component.go
```diff
@@ -771,6 +789,8 @@ type RoleSelectComponent struct {
 	ValueLimits [2]int `json:"-"`
 	// Disabled disables the select if true.
 	Disabled bool `json:"disabled,omitempty"`
+	// SelectedRoles is the slice of RoleIDs that are marked as selected by default
+	SelectedRoles []RoleID `json:"-"`
```
diamondburned: See [1].

discord/component.go
```diff
@@ -875,6 +937,8 @@ type ChannelSelectComponent struct {
 	Disabled bool `json:"disabled,omitempty"`
 	// ChannelTypes is the types of channels that can be chosen from.
 	ChannelTypes []ChannelType `json:"channel_types,omitempty"`
+	// SelectedChannels is the list of channels that are marked as selected by default.
+	SelectedChannels []ChannelID `json:"-"`
```
diamondburned: See [1].

discord/component.go
```diff
@@ -822,6 +858,9 @@ type MentionableSelectComponent struct {
 	ValueLimits [2]int `json:"-"`
 	// Disabled disables the select if true.
 	Disabled bool `json:"disabled,omitempty"`
+	// SelectedMentions is the slice of discord.UserID's and discord.RoleID's 
+	// that are marked as selected by default
+	SelectedMentions []interface{} `json:"-"`
```
diamondburned: If we're using a hidden field, then it's best to just have separate hidden `DefaultUsers` and `DefaultRoles` fields, then combine them in `MarshalJSON`.

diamondburned (changes requested):

discord/component.go
MCausc78: This breaks proper order. There might be `{(user, 2), (role, 3), (user, 1)}` sequence.

CosmicPredator: I agree with @MCausc78. What's wrong with using plain interfaces? It's already documented above the field for the users. 

I can add an example snippet in the comments, If it's not clear enough

diamondburned: > What's wrong with using plain interfaces? It's already documented above the field for the users.

It's not type-safe.

The only way I would agree to moving forward with this solution is to make custom `SelectedUserID` and `SelectedRoleID` types and a `SelectableID` union type to combine the two. Arikawa tries its best to not have any ambiguous/error-prone `any` types in its code, so this shouldn't be an exception.

CosmicPredator: @diamondburned How about this implementation?

```go
type DefaultMention struct {
	userId *UserID `json:"-"`
	roleId *RoleID `json:"-"`
}

// Creates new DefaultMention type with only UserID
func UserMention (userId UserID) DefaultMention {
	return DefaultMention{userId: &userId}
}

// Creates new DefaultMention type with only RoleID
func RoleMention(roleId RoleID) DefaultMention {
	return DefaultMention{roleId: &roleId}
}
```

and when implementing,

```go
DefaultMentions: []discord.DefaultMention{ discord.UserMention(793688107077468171), discord.RoleMention(1299037008332849212) },
```

diamondburned: I'm not sure about the naming, but this looks good. You can use `{UserID, RoleID}` for the types instead of `{*UserID, *RoleID}` then check for `IsValid()`.

CosmicPredator: Commit done...! and for the namings, we can change it if the users are feeling discomfort with it.

CosmicPredator:
@diamondburned I guess I updated everything and you are gtg

discord/component.go
```diff
@@ -847,6 +847,22 @@ func (s *RoleSelectComponent) MarshalJSON() ([]byte, error) {
...
 }
 
+// DefaultMention type is a Union type which packs both UserID and RoleID
+type DefaultMention struct {
+	userId UserID `json:"-"`
+	roleId RoleID `json:"-"`
+}
+
+// Creates new DefaultMention type with only UserID
+func UserMention (userId UserID) DefaultMention {
```
diamondburned: I think a better name for this would be `DefaultUserMention()`.

diamondburned: Documentation should be in the format

```go
// DefaultUserMention creates...
```

discord/component.go
```diff
@@ -858,11 +874,10 @@ type MentionableSelectComponent struct {
...
 	// Disabled disables the select if true.
 	Disabled bool `json:"disabled,omitempty"`
-	// DefaultMentions is the slice of discord.UserID's and discord.RoleID's 
-	// that are marked as selected by default
-	// Example: 
-	// DefaultMentions: []interface{}{ discord.RoleID(78402180381208302), discord.UserID(87028080234556) }
-	DefaultMentions []interface{} `json:"-"`
+	// DefaultMentions is the slice of User / Role Mentions that are selected by default
+	// Example:
+	// DefaultMentions: []DefaultMention{ NewUserMention(0382080830233), NewRoleMention(4820380382080) }
```
diamondburned: Fix example code.

```go
// Example:
//   DefaultMentions: []DefaultMention{
//     DefaultUserMention(userID1),
//     DefaultRoleMention(roleID2),
//   }
```

diamondburned (changes requested):

CosmicPredator:
@diamondburned Fixed the requested

CosmicPredator:
@diamondburned Is there anything that needs to be fixed still ??

diamondburned:
Thanks for the bump! I just forgot about this PR for a bit.

diamondburned (approved):

PR #469, opened 2025-03-17
api: Support with_counts for Guilds method

twoscott:
Supports missing `with_counts` field for the `/users/@me/guilds` endpoint and adds helper methods to make fetching all guilds and members clearer. Also updates the `MaxGuildFetchLimit` constant to 200 as per [Discord API documentation](https://discord.com/developers/docs/resources/user#get-current-user-guilds).

### DIFF
--- a/api/guild.go
+++ b/api/guild.go
@@ -12,7 +12,8 @@ import (
 
 // MaxGuildFetchLimit is the limit of max guilds per request, as imposed by
 // Discord.
-const MaxGuildFetchLimit = 100
+// https://discord.com/developers/docs/resources/user#get-current-user-guilds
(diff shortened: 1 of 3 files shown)


diamondburned (dismissed):
This introduces the new `All...` convention, but it looks fine to me.

Minor suggestion: I would create a `GuildsBeforeData` that contains a `WithCounts` rather than having a new positional `withCounts` parameter, since it's optional.

twoscott:
> Minor suggestion: I would create a `GuildsBeforeData` that contains a `WithCounts` rather than having a new positional `withCounts` parameter, since it's optional.

Good idea, added this in 5a6b615

diamondburned (approved):

PR #470, opened 2025-03-21
Respect `retry_after` in HTTP 429 response

merlinfuchs:
Arikawa does prevent most 429's by respecting the headers returned by Discord but some rate limits aren't indicated by the headers.

### DIFF
--- a/utils/httputil/client.go
+++ b/utils/httputil/client.go
@@ -17,6 +17,10 @@ import (
 // StatusTooManyRequests is the HTTP status code discord sends on rate-limiting.
 const StatusTooManyRequests = 429
 
+// StatusNirnTimeout is the HTTP status code Nirn Proxy sends when the request
+// times out.
+const StatusNirnTimeout = 408
+
 // Retries is the default attempts to retry if the API returns an error before
 // giving up. If the value is smaller than 1, then requests will retry forever.
 var Retries uint = 5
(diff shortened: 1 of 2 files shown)


utils/httputil/client.go
```diff
@@ -230,10 +234,34 @@ func (c *Client) request(
 			continue
 		}
 
-		if status = r.GetStatus(); status == StatusTooManyRequests || status >= 500 {
+		status = r.GetStatus()
+
+		if status == StatusNirnTimeout || status >= 500 {
```
diamondburned, line 239: In the case of a 5xx response, I think we should do our own exponential backoff to prevent spamming Discord.

diamondburned, line 239: Rather than adding a `StatusNirnTimeout`, I think we should just write `status == http.StatusRequestTimeout`.

utils/httputil/client.go
```diff
@@ -230,10 +234,34 @@ func (c *Client) request(
...
+			buf.ReadFrom(body)
+
+			var errBody struct {
+				RetryAfter float32 `json:"retry_after"`
+			}
+
+			err := json.Unmarshal(buf.Bytes(), &errBody)
+
+			if err == nil {
+				time.Sleep(time.Duration(errBody.RetryAfter) * time.Second)
```
diamondburned, line 257: `time.Sleep` is not context-aware. Prefer writing:

```go
select {
case <-ctx.Done():
    doErr = ctx.Err()
    continue // or break out of the loop, but this will work too
case <-time.After(...):
    continue
}
```

utils/httputil/client.go
```diff
@@ -230,10 +234,34 @@ func (c *Client) request(
...
+			defer body.Close()
+
+			buf := bytes.Buffer{}
+			buf.ReadFrom(body)
+
+			var errBody struct {
+				RetryAfter float32 `json:"retry_after"`
+			}
+
+			err := json.Unmarshal(buf.Bytes(), &errBody)
```
diamondburned, line 254: We can shorten this to:

```suggestion
			err := json.NewDecoder(body).Decode(&errBody)
```

utils/httputil/client.go
```diff
@@ -230,10 +234,34 @@ func (c *Client) request(
...
+
+			if err == nil {
```
diamondburned, line 256: Happy path: put the `if err != nil` check first, then put the other case after the check.

diamondburned (changes requested):

merlinfuchs:
I mainly made these changes so I can it in a project, I agree with your comments and can make the changes when I get around to it! Happy for you to also close this PR and make the changes :)

PR #472, opened 2025-03-26
Support GIF banner URLs

twoscott:
Currently, `Guild.BannerURL()` returns the banner as a `PNG` URL no matter the circumstances. I'm not sure if banners were PNG-only when they were first added to Discord, but they can be animated now and we can use the same logic to determine if it's animated or not as we do with user avatars and guild icons.

### DIFF
--- a/discord/guild.go
+++ b/discord/guild.go
@@ -142,15 +142,16 @@ func (g Guild) IconURLWithType(t ImageType) string {
 }
 
 // BannerURL returns the URL to the banner, which is the image on top of the
-// channels list. This will always return a link to a PNG file.
+// channels list. Auto detects a suitable image type. An empty string is
+// returned if the guild has no banner.
 func (g Guild) BannerURL() string {
-	return g.BannerURLWithType(PNGImage)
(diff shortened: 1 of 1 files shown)


diamondburned:
> I'm not sure if banners were PNG-only when they were first added to Discord

I think they are. I'm not sure if I would agree with this change, since it is an unnecessary breaking change: clients that don't support GIFs will now unexpectedly get a GIF sometimes.

twoscott:
> > I'm not sure if banners were PNG-only when they were first added to Discord
> 
> I think they are. I'm not sure if I would agree with this change, since it is an unnecessary breaking change: clients that don't support GIFs will now unexpectedly get a GIF sometimes.

That's fair, I'd think they'd use `Guild.BannerURLWithType()` if they wanted to ensure it's a specific type but I suppose this has been documented to return PNG only. We could add another method then perhaps? `BannerURLAuto()` or something along those lines? Obviously `...WithType(AutoImage)` could be used but the available methods could lead someone to believe GIF banners aren't supported.

PR #483, opened 2025-09-02
feat(gateway): add ability to set custom identify properties

ayn2op:


gateway/identify.go
```diff
@@ -96,12 +96,62 @@ func (id *Identifier) QueryGateway(ctx context.Context) (gatewayURL string, err
...
+	_ IdentifyProperties = BotIdentifyProperties{}
+	_ IdentifyProperties = UserIdentifyProperties{}
+)
+
+type (
+	IdentifyProperties interface {
+		isIdentifyProperties()
+	}
+
+	BotIdentifyProperties struct {
```
diamondburned: imo maybe call this `BasicIdentifyProperties`? I don't think the difference means too much (e.g. if a bot has extra attributes, Discord isn't going to bitch about it), but Dissent uses these 3 + only a few more and it's going fine.

gateway/identify.go
```diff
@@ -96,12 +96,62 @@ func (id *Identifier) QueryGateway(ctx context.Context) (gatewayURL string, err
...
+		// OS is the operating system of the client.
+		OS string `json:"os"`
+		// Browser is the browser the client is using.
+		Browser string `json:"browser"`
+		// Device is the model of the mobile device the client is running on.
+		Device string `json:"device"`
+	}
+
+	UserIdentifyProperties struct {
```
diamondburned: I still think a map is better here:

```go
type CustomIdentifyProperties map[IdentifyPropertyKey]any

type IdentifyPropertyKey string

const (
	IdentifyOS      IdentifyPropertyKey = "os"
	IdentifyBrowser IdentifyPropertyKey = "browser"
	IdentifyDevice  IdentifyPropertyKey = "device"
// add the rest
)
```

Maybe keep only these 3 keys, or maybe add all the known ones? But at least this means we don't need to actively catch up with Discord upstream's arbitrary fields.

gateway/identify.go
```diff
@@ -96,12 +96,62 @@ func (id *Identifier) QueryGateway(ctx context.Context) (gatewayURL string, err
...
+		ReferringDomainCurrent string `json:"referring_domain_current,omitempty"`
+	}
+)
+
+func (b BotIdentifyProperties) isIdentifyProperties()  {}
+func (u UserIdentifyProperties) isIdentifyProperties() {}
+
 // IdentifyCommand is a command for Op 2. It is the struct for a data that's
 // sent over in an Identify command.
 type IdentifyCommand struct {
```
diamondburned, line 118: Add an Unmarshaler here to ensure IdentifyProperties gets unmarshaled correctly. Probably by ensuring that `BotIdentifyProperties` is used if the keys are exactly `{"os", "browser", "device"}` (package maps and slices extremely helpful here) otherwise `CustomIdentifyProperties` is used.

diamondburned (changes requested):

diamondburned (approved):

PR #487, opened 2025-09-22
feat(httputil): add NewClientWithDriver function

ayn2op:
The name feels a bit weird, open for suggestions. Perhaps, we should get rid of one of the intermediates.

### DIFF
--- a/utils/httputil/client.go
+++ b/utils/httputil/client.go
@@ -44,8 +44,12 @@ type Client struct {
 }
 
 func NewClient() *Client {
+	return NewFromDriverClient(httpdriver.NewClient())
+}
+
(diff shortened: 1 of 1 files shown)


utils/httputil/client.go
```diff
@@ -44,8 +44,12 @@ type Client struct {
 }
 
 func NewClient() *Client {
+	return NewFromDriverClient(httpdriver.NewClient())
+}
+
+func NewFromDriverClient(driver httpdriver.Client) *Client {
```
diamondburned: `NewClientWithDriver` was a pretty good option.

diamondburned (changes requested):

ayn2op:
Done!

diamondburned (approved):

PR #499, opened 2026-02-13
feat(message): Add MessageCall object

xqrs:
you forgot to add it https://docs.discord.food/resources/message#message-call-object

### DIFF
--- a/discord/message.go
+++ b/discord/message.go
@@ -112,7 +112,7 @@ type Message struct {
 	// [MessageReferenceTypeForward].
 	MessageSnapshots []MessageSnapshot `json:"message_snapshots,omitempty"`
 
-	// Call object
+	// Call is the private channel call (MessageCall) that prompted this message.
 	Call MessageCall `json:"call,omitempty"`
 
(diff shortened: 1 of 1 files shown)


discord/message.go
```diff
@@ -112,6 +112,9 @@ type Message struct {
 	// [MessageReferenceTypeForward].
 	MessageSnapshots []MessageSnapshot `json:"message_snapshots,omitempty"`
 
+	// Call object
```
ayn2op:
```suggestion
	// Call is the private channel call (MessageCall) that prompted this message.
```

discord/message.go
```diff
@@ -120,6 +123,13 @@ type Message struct {
 	Stickers []StickerItem `json:"sticker_items,omitempty"`
 }
 
+type MessageCall struct {
+	// Channel UserIDs who participated in the call.
```
ayn2op: The comment for this field could be improved.
```suggestion
	// Participants is the IDs of the users who participated in the call.
```

ayn2op (approved):

PR #518, opened 2026-03-05
discord: rename CheckboxGroupComponent Value to Values with type []string

ayn2op:
https://docs.discord.com/developers/components/reference#checkbox-group-structure

### DIFF
--- a/discord/component.go
+++ b/discord/component.go
@@ -1649,7 +1649,7 @@ type CheckboxGroupComponent struct {
 	// Whether the file upload requires files to be uploaded before submitting the modal (defaults to `true`)
 	Required bool `json:"required,omitempty"`
 	// The values of the selected options.
-	Values []string `json:"value,omitempty"`
+	Values []string `json:"values,omitempty"`
 }
 
 // Type implements the Component interface.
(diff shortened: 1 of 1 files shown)


discord/component.go
```diff
@@ -1648,8 +1648,6 @@ type CheckboxGroupComponent struct {
 	ValueLimits [2]int `json:"-"`
 	// Whether the file upload requires files to be uploaded before submitting the modal (defaults to `true`)
 	Required bool `json:"required,omitempty"`
-	// The value of the selected option, or null if no option is selected
-	Value bool `json:"value,omitempty"`
```
topi314, line 1652: this is needed since the components are also used to receive interactions
but it should be:
```
Values []string `json:"value,omitempty"`
```

See: https://docs.discord.com/developers/components/reference#checkbox-group-interaction-response-structure

discord/component.go
```diff
@@ -1648,8 +1648,8 @@ type CheckboxGroupComponent struct {
...
+	// The values of the selected options.
+	Values []string `json:"value,omitempty"`
```
topi314:
```suggestion
	Values []string `json:"values,omitempty"`
```

diamondburned (approved):

PR #519, opened 2026-03-05
refactor: run go fix ./...

ayn2op:


discord/emoji.go
```diff
@@ -21,7 +21,7 @@ type Emoji struct {
 	// User is the user that created the emoji.
 	//
 	// This field is only available for custom emojis.
-	User User `json:"user,omitempty"`
+	User User `json:"user"`
```
diamondburned: What's with removing `omitempty`?

diamondburned (changes requested):

discord/emoji.go
ayn2op: As far as I understand, since omitempty has no effect on struct-typed fields, `omitempty` becomes redundant on these fields and has no effect. We have two options here, either remove omitempty (same behavior as now since `omitempty` has no effect anyway) or replace `omitempty` with `omitzero`.

From `go tool fix help omitzero`:

omitzero: suggest replacing omitempty with omitzero for struct fields

The omitzero analyzer identifies uses of the `omitempty` JSON struct
tag on fields that are themselves structs. For struct-typed fields,
the `omitempty` tag has no effect on the behavior of json.Marshal and
json.Unmarshal. The analyzer offers two suggestions: either remove the
tag, or replace it with `omitzero` (added in Go 1.24), which correctly
omits the field if the struct value is zero.

However, some other serialization packages (notably kubebuilder, see
https://book.kubebuilder.io/reference/markers.html) may have their own
interpretation of the `json:",omitzero"` tag, so removing it may affect
program behavior. For this reason, the omitzero modernizer will not
make changes in any package that contains +kubebuilder annotations.

Replacing `omitempty` with `omitzero` is a change in behavior. The
original code would always encode the struct field, whereas the
modified code will omit it if it is a zero-value.

diamondburned: I think we should replace with `omitzero`. `encoding/json` is really the only encoder that we support right now.

diamondburned (approved):

PR #522, opened 2026-03-09
Make textinputcomponent's label omitempty

hi117:
This is for when its used as part of a label component


ayn2op (approved):
Note to self: remove the label field entirely

https://docs.discord.com/developers/components/reference#text-input

